PR #19, opened 2021-04-15
Add web app pages

apt105:
Can be seen at https://meshtastic-7p5lrdwty-apt105.vercel.app/docs/software/web/web-app-software


sachaw:
Looks good, just a few changes:
Make the section just called `Web Interface`, because the repo/project that's the main web client has changed a few times, and meshtastic-web is outdated (meshtastic-client is the current) , that way it's client agnostic.
hence you can remove the compatibility note as the current client works fine with 1.2x.

I'll leave this open, because I'll rewrite a lot of the content, and all of the development notes will be moved into the development /meshtastic.js section

apt105:
Suggested changes made, looking forward to seeing the new client!

PR #20, opened 2021-04-15
Update index.tsx

apt105:
Quick attempt to improve the landing page now that we've gone live. Can update further later, but wanted a bit more information to be available for first time visitors.

Example at https://meshtastic-r2hyw0nsd-apt105.vercel.app/

### DIFF
--- a/website/src/pages/index.tsx
+++ b/website/src/pages/index.tsx
@@ -14,10 +14,7 @@ const features = [
     imageUrl: "img/software.svg",
     description: (
       <>
-        Having an ever-growing, vibrant software ecosystem, there's almost
-        certainly, a capable solution ready to go for your project. With
-        libraries for Typescript and Python, apps for all major platforms and a
-        whole suite of native integrations, there's something for everyone.
+        With an ever-growing array of applications, plugins and interface options, Meshtastic can be used with a variety of different devices to create a network that suits your requirements.
       </>
     ),
   },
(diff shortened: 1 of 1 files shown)


sachaw:
I'm all for adding more info to the landing page, but it needs to be flashier and look more formatted than that,
if you want some inspiration, take a look at:
https://github.com/ (incognito)
https://www.prisma.io/
https://nextjs.org/

apt105:
Fully appreciate that, but with the site going live I want to get just something on there to give a bit of explanation, and also remove the boilerplate from docusaurus under the first icon.

jfirwin:
I think we could replace your bullet points with icons like the front page has. They look like they are pretty responsive and with the right images it could really pop.

I'd also eventually like those to link to the `About` page or sidebar if we break that section up much.

See https://meshtastic.letstalkthis.com/

I'm all for doing some image designing! Let me know if you want me to take a stab at it.

That way it covers the flashy aspect and is more pertinent info than the stock docusaurus layout.

apt105:
I currently have no preference as to what goes on the front page. It would be nice to get something flashy as per Sacha's links, but my main concern is that meshtastic.org is now showing a front page which explains nothing about what Meshtastic is, and references an entirely different project in the information it does give. My PR is a quick fix to remedy this in the full knowledge that it needs replacing with something substantially better in the future. If one of you can put something better together today or tomorrow, great, lets go with that instead, but if it's going to be any longer, then we need to just put something there instead of what we already have.

jfirwin:
I'll try making some artwork for those bullet points you added.

PR #22, opened 2021-04-17
Add Legal section

jfirwin:
Definitely needs more TLC, but until we have someone review this we want to have something stating terms of use and specifying the trademark.

I suggested in slack that we should consider using [Raspberry Pi Foundation's page](https://www.raspberrypi.org/trademark-rules/) as an example of a good trademark page.


jfirwin:
available for viewing [here](https://meshtastic-git-legal-jfirwin.vercel.app/)

sachaw:
I'd put a "Legal" link in the footer under Other

jfirwin:
As an additional spot or in place of the sidebar?

sachaw:
Under here https://github.com/meshtastic/Meshtastic/blob/912213f1f147722009da5df91ca483dbc7bd8edb/website/docusaurus.config.js#L98

jfirwin:
This is done. Viewable [here](https://meshtastic-git-legal-jfirwin.vercel.app/)

I added an overview page (empty) to be filled out once we get more content/decide if we are doing a contributor agreement, etc.

apt105:
Looks good, don't know if you want to add the disclaimer on software patents that was up previously. I can't find it at the moment...

jfirwin:
If we find it as we are clearing things out, I think that's a perfect place to put this. This section is definitely incomplete, and I'm not a lawyer so I'm probably not the right person to fill it all out. I'll keep adding bits and pieces in the future.

apt105:
Found it

jfirwin:
Awesome! Thanks! I'll move that over tonight or tomorrow.

PR #25, opened 2021-04-19
Hardware

apt105:
I've added in detail about all the modules and made the purchase link more evident. See https://meshtastic-7s86x48yg-apt105.vercel.app/docs/hardware/tbeam-hardware

### DIFF
--- a/website/docs/hardware/heltec.md
+++ b/website/docs/hardware/heltec.md
@@ -18,6 +18,7 @@ sidebar_label: Heltec
 * Reset and Program switches
 * No GPS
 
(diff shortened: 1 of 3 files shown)


jfirwin:
This looks awesome! That's a lot of work cataloging all of that. I was wondering if banggood/Amazon links would also be included down the line? If so, we may want images for each respective marketplace. But I think that's a "down the line" issue.

apt105:
I'll try to get round to alternative links at some point, but going to see what else I can do for the time being.

PR #32, opened 2021-05-01
Community projects

apt105:
Some pages for the projects that have been done to work with Meshtastic but which are not part of the Meshtastic project. 
Permission has been sought and granted by both PaulMandel and ZebusJesus, who are both happy with the info available on the pages. Example page: https://meshtastic-git-community-projects-apt105.vercel.app/docs/software/community/community-overview


sachaw:
CI failing wasn't this, I'd not setup the action correctly.

sachaw:
I'd add an admonition on each project page saying something along the lines of:
```
This is a community project maintained by xxx
Development can be followed here xxx
Support should be sought from the respective authors.
```
or something like that.

apt105:
Good point, I've added the admonitions.

PR #33, opened 2021-05-01
Device software pages

apt105:
Includes overview, multichannel, admin channel, GPIO, power state information, and how to set wifi. Example: https://meshtastic-git-devices-apt105.vercel.app/docs/software/device/device-software


apt105:
I've tested most of the documentation apart from the GPIO stuff, I don't fancy frying my boards and I don't really know enough about which GPIOs would be safe to test on which boards.

PR #36, opened 2021-05-04
Settings

jfirwin:
Edit: Available to view [here](https://meshtastic-git-settings-jfirwin.vercel.app/docs/settings).

I'm ready for a review. Bring on the good, bad, and ugly. Some notes:

- All settings that are set using `meshtastic --set` should be accounted for, but there are other settings that are missing which use different flags. Those will need to be added.
- Initially I had each setting's description in the table, but later moved those out of the table so that more "long-form" descriptions can be made and admonitions can be added for specific settings
- There are a lot of `<!--- TODO --->` comments, I commented them out so they don't show up in production, but we will be able to see them and not forget them.
- Styleguide: Each page has the following
  - Overview
  - Settings
    - Settings table
    - H3 Setting descriptions
  - Details (for any explanation)
    - CLI examples (I'm not opposed to moving these to an H2 heading, just hadn't thought of it through my iterations of formatting this)

Please review the content too, some things I guessed or tried to extrapolate from the `meshtastic/Meshtastic-device` repo, and I'm no python expert so I may be dead wrong on a few things.


jfirwin:
Also, I was able to build this locally, are we still having issues with the pull_request check? I may be missing something. I'll dig deeper.

sachaw:
Yeah, still haven't seemed to be able to nail it, should fix it soon

sachaw:
Look god, only thing is, I think that this should be located under software, possibly in the future this can become the python docs, when we have more cli's

jfirwin:
Yeah, @apt105 brought that up too, I wasn't sure where to put it. I hope to detail every setting that can be set on the device all in one spot. The CLI makes sense, I just want the user to have it clearly marked as something to the effect of "All device settings" so it's not nested (too deep) somewhere. In the future we could add sections for changing settings on Android/iOS/Web etc. I was thinking this could be possible using [tabs](https://docusaurus.io/docs/markdown-features/tabs). Each platform setting example could be showcased in its own tab.

jfirwin:
I'll move this to `Software > Settings` and test out the Tab functionality on the examples.

jfirwin:
I'm really liking how this is coming together with the tabs. It does make the documents a little messier to look at with the JSX in there, but on the finished site I think it looks really good. It'll also allow us to put setting details for iOS, web, etc without bogging down the page. Also, I've grouped them all so if the user has clicked on `CLI` (default) all examples show using the CLI, if I click on one page to android, all the pages change to show android instructions since its a global variable.

Speaking of android, I don't have one so those instructions I'll leave to someone who can check them!

sachaw:
Looking great, let me know when you want me to merge even if it's wip, just so it doesn't get unwieldly

jfirwin:
I'm wondering it it would work better under `Hardware > Settings`. But other than that I'm fine to merge.

My thought is that it is device configuration related and that way it won't get tied in with the other software.￼

PR #38, opened 2021-05-09
Antenna

apt105:
An update to the section on antennas. https://meshtastic-git-antenna-apt105.vercel.app/docs/hardware/antenna


jfirwin:
I really like this as a section, it is a huge topic so having this split up will make it easier to digest. Great work!

sachaw:
Looks great!

PR #40, opened 2021-05-16
Firmware details

jfirwin:
added admonitions about tbeam0.7 mentioned on slack


sachaw:
Let me know when your ready to merge!

jfirwin:
I think I'm set. I added the same link that's on the firmware page to the `getting started/overview` page but it was breaking links and I'm not sure why so I just removed it altogether.

Any idea what was different? Now I'm thinking it may have been the fact overview has a set url?

sachaw:
I'll look into it later. As long as it builds it's good

PR #42, opened 2021-05-21
Attempting to clarify what send_owner_interval is

thebentern:
My notes are based on this information: https://meshtastic.discourse.group/t/real-time-communication-and-location-update/1990/2

### DIFF
--- a/website/docs/software/settings/power.md
+++ b/website/docs/software/settings/power.md
@@ -25,7 +25,7 @@ import TabItem from '@theme/TabItem';
 | phone_timeout_secs | `integer` (seconds) | `0` |
 | screen_on_secs | `integer` (seconds) | `0` |
 | sds_secs | `integer` (seconds) | `0` |
-| send_owner_interval | `integer` | `4` |
+| send_owner_interval | `integer` (sent every x network pings) | `4` |
 | wait_bluetooth_secs | `integer` (seconds) | `0` |
 
 :::note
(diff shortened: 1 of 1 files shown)


jfirwin:
This looks great. I'm waffling back and forth on wether we should include `(pings)` in the table where you removed seconds. Any thoughts? I'm ok either way, as it's described below in the setting section

thebentern:
Maybe (per network ping). 
I'd like to actually define what a ping is in some section of the docs. It's another one of those ambiguous terms that gets thrown around a lot but is entirely different in the context of the meshtastic mesh behavior. I _think_ it means position broadcasts and owner info broadcasts, but I can't say for certain.

jfirwin:
Maybe `(network ping interval)` or `(each x/interval pings)`?

I agree that pings should be more well defined. We can add a  html commented out TODO mentioning to link to that.
```
<!--- TODO add link to ping explanation --->
```

thebentern:
> Maybe `(network ping interval)` or `(each x/interval pings)`?
> 
> I agree that pings should be more well defined. We can add a html commented out TODO mentioning to link to that.
> 
> ```
> 
> ```

Maybe send per `(sent per x network pings)`

jfirwin:
`(sent every x network pings)`? I think every might be a little more clear than per, but I think you're right. I'm good with either.

thebentern:
> `(sent every x network pings)`? I think every might be a little more clear than per, but I think you're right. I'm good with either.
Perfect!

sachaw:
Ready to merge?

jfirwin:
I believe so! I've been on mobile all weekend so I haven't been able to "approve" it.

PR #43, opened 2021-05-22
esptool command line

michelepagot:
At least on Windows10, the esptool cannot be accessed at "esptool.py" but has to be accessed on "esptool"

### DIFF
--- a/website/docs/getting-started/flashing-firmware.md
+++ b/website/docs/getting-started/flashing-firmware.md
@@ -185,11 +185,11 @@ pip --version
 Connect the radio to your computer using a data USB cable. Confirm your device is talking to your computer using the following command:
 
   ```bash title="Command"
-  esptool.py chip_id
+  esptool chip_id
   ```
   ```bash title="Expected Output"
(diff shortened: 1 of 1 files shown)


sachaw:
Good catch!

jfirwin:
Is that assuming the user is using [GitBash for windows](https://meshtastic.org/docs/getting-started/flashing-firmware#install-prerequisite-software)?

I'll need to spin up a virtual machine to test it, but the instructions were written under that assumption. If so, good catch! I must have been a little copy/paste happy writing it up between each OS. If not, we may want to add an admonition to the windows sections stating what flavor command line the instructions assume.

michelepagot:
I was about to also propose a device-update.bat and device-install.bat in Meshastic-Device repo, to remove GitBash dependency at all. I'll let you know if it is feasable

jfirwin:
If that ends up working it would be great! We can adjust the Windows sections accordingly.

Does the meshtastic CLI tool operate the same on windows when not using gitbash? If it's different I may need to adjust the settings pages tabs as well.

PR #76, opened 2021-10-26
Improve documentation

joarSv:
No documentation on usage or configuration is currently written. This PR provides a good starting point.


logiota:
it also costs 10x more with lower precision :)

garthvh:
All sensors now use I2C.

PR #126, opened 2021-12-01
Update the project description

garthvh:
Deemphasize optional GPS features, add group size limits and call out that each device can only support one connection at a time.


sachaw:
Good points

PR #170, opened 2022-01-04
Broken Link Fixes in Docs

dfloer:
Found a few broken/outdated links in the docs and fixed them.

For the two links to ITU docs (1ecafbb), they seem to have been moved, so I linked to the Internet Archive versions. This may be a good idea to do for external links to content that's expected to remain static.


mkinney:
Thanks for those fixes/changes

PR #184, opened 2022-01-19
reorganize navigation and sidebars

jfirwin:
I'm open to any and all comments. I know there will be some things that need to be reworked on the document level with this new navigation layout. So I think we can fix those things with future PRs.


garthvh:
Looks good, makes a lot more sense

PR #190, opened 2022-01-22
add more info about env sensors

mkinney:


garthvh:
While gas is funny, the main use of that sensor is AQI

PR #192, opened 2022-01-22
add quotes around cli commands

mkinney:


jfirwin:
I'm on mobile, but the "template" for the plug-in settings is probably the range test plugin document.

I'll try to add some headers between these commands when I get a chance.

PR #218, opened 2022-01-31
Update build.md

RicInNewMexico:
Added python3-dev to list of required packages for linux build environment

### DIFF
--- a/docs/developers/build.md
+++ b/docs/developers/build.md
@@ -31,7 +31,7 @@ To get a clean build you may have to delete the auto-generated file `./.vscode/c
 
 ```
 sudo apt-get update
-sudo apt-get install python3 python3 g++ zip
+sudo apt-get install python3 python3-dev g++ zip
 ```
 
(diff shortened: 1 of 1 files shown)


docs/developers/build.md
```diff
@@ -31,7 +31,7 @@ To get a clean build you may have to delete the auto-generated file `./.vscode/c
 
 ```
 sudo apt-get update
-sudo apt-get install python3 g++ zip
+sudo apt-get install python3 python3 g++ zip
```
garthvh, line 34: Why is python3 here twice?

RicInNewMexico: That should be python3-dev and I've corrected that in my repo. I'm not sure why the -dev got dropped when I saved it from VStudio.. 

Best next step? cancel this pull and generate new?

RicInNewMexico: Actually it looks like that won't be necessary. If you refer to the newest commit it corrects it to python3-dev

PR #237, opened 2022-02-08
Better compress images for speed/size

zjwhitehead:
I noticed some slow loading pages particularly while browsing the docs. 
It looks like the raw images are being served without proper sizing or compression. 
This is a first pass at simply better compressing the images and doesn't impact the dimensions or quality. Further updates can be made to use improved image formats (eg .webp) and properly size the images.

As an example ([page](https://meshtastic.org/docs/hardware/supported/techo)) we are loading a 13.1MB image which took over 10 seconds on a high speed connection.

image

- Run using ImageOptim 1.8.9 using lossless compression
On average we are saving more than 26% in file sizes

image


jfirwin:
Great work. Optimization is definitely something that has not been on the road map to date.

PR #240, opened 2022-02-12
Major cleanup to mesh-alg doc.

dfloer:
Vercel preview of the proposed changes: https://meshtastic-git-master-dfloer.vercel.app/

Biggest change was that I removed the notes section. It seems like it's mostly/entirely obsolete, and not related to the docs page its on.

Also did some formatting tweaks (made heading Title Case), moved around and reworded things, and added a couple tables to better convey some of the things being discussed. Plus, I at least, find it easier to the layout of things like a packet structure as a table, rather than words. This also meant moving the notes out of the table, and I'm not 100% sure I like this style.

One of the tables I added has calculations from the code on the minimum times it takes to transmit packet, at the extremes of the range.

I added a "Layer 0" and moved the discussion of the preamble there, because it seemed more pertinent there than in the packet structure, as it _isn't_ encoded at that level, but appears to be only at the symbol level.

I'm not sure if the section on lines 81-84 is accurate anymore. I couldn't find the comment in the current version of the file.

### DIFF
--- a/docs/developers/device/mesh-alg.md
+++ b/docs/developers/device/mesh-alg.md
@@ -4,219 +4,103 @@ title: Mesh broadcast algorithm
 sidebar_label: Mesh algorithm
 ---
 
-## Current algorithm
+## Current Algorithm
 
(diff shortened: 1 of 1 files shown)


jfirwin:
Great work. This is a quality, in depth, readable document now!

PR #251, opened 2022-02-21
changed #8c8ea0 to #4d4d4d

jfirwin:
The purple-ish hue to the current icons on the front page have always felt like they clash with the theme a little bit. I used the light mode/dark mode toggle button's background color and replaced the colors.

I also thinned the stroke on the puzzle piece and the key to make them look a little less clunky.

Available for preview [here](https://meshtastic-git-colors-jfirwin.vercel.app).


sachaw:
Looks good

PR #253, opened 2022-02-21
move settings folder

jfirwin:
This commit is moving settings out of the software folder, so settings will reside at `docs/settings/xyz` instead of `docs/software/settings/xyz`

This will break links that are in the forum or discord for the mean time. I'm hoping not to break too many links, but we decided previously not to deal with redirects until we are closer to a "final form" with the docs.


jfirwin:
I sent out a message on discord and the forum alerting people of the possibility of broken links. Let's break some stuff.

PR #265, opened 2022-02-26
Typo

jfirwin:


jfirwin:
@mkinney elsewhere (manual firmware flashing pages) we list Linux, macOS, and windows as options. These variables are global and should select across the whole site provided they are the same everywhere, should we change Ubuntu to Linux or Linux to ubuntu so those operating systems line up everywhere?

PR #282, opened 2022-03-17
Mention that the serial module isn't yet implemented

raldi:
And add some tips about pins.


raldi:
Ah, just got some clarification on Discourse. Editing...

jfirwin:
I was just looking at this and the discourse messages. Sounds good. Once you've made the edits I'll merge it in.

PR #288, opened 2022-03-18
Updating docs to correct syntax

ayysasha:
Syntax correct from xxx_module_xxx to xxx_plugin_xxx


caveman99:
That reverts the docs from 1.3 to 1.2 syntax. We'd rather put up a banner to aadvise people of the different command styles.

PR #289, opened 2022-03-18
include a notice about naming change in plugins to modules (1.3 vs. 1.2)

caveman99:


jfirwin:
Amazing. This was on my list of things for today, so you just saved me a bunch of time. Also great use of the partial markdown files!

caveman99:
it would even be better if i could have made the example from https://docusaurus.io/docs/markdown-features/react#importing-markdown work like they advertise it. i couldn't for the life of me replace the curly braces like they do. if that worked, you could reduce that snippet to one and mention the actual module name more clearly

jfirwin:
I know that the compiler struggles a bit with anything inside of the admonitions unless there is an empty new line after `:::x` and before `:::`

caveman99:
not newline, but you had to wrap it in '<span>'. PR update coming up.

jfirwin:
Interesting, good to know! I'll hold on merging.

caveman99:
looking good. ship it :-)

PR #291, opened 2022-03-21
Add battery purchase instructions

jberenhaus:
Added instructions on how to purchase a battery and some preferences

### DIFF
--- a/docs/getting-started/getting_started.md
+++ b/docs/getting-started/getting_started.md
@@ -33,7 +33,7 @@ The easiest way is to [buy a device with the software already installed](https:/
 
 #### Purchase battery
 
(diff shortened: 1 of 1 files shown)


sachaw:
I'd change the wording, as we have many people who reside outside the US.

Make it just an option instead of a recommendation.

jfirwin:
I was thinking along the same lines. If we get that changed we can merge it in.

Thanks for your contribution!

jberenhaus:
Oh good call, I updated it to say "One of our favorite places for folks in the US to buy the necessary 18650 Li-Ion battery is..."

Let me know if there's any favorite places to buy it outside the US you want me to add or any other wording recommendations :)

sachaw:
Tweaked it a little more, now nobody can complain if they don't get the results they expected.

Thanks for the pr

PR #294, opened 2022-03-22
De-duping and ordering T-Beam hardware listings

jberenhaus:
V1.1 and Meshtastic had the same link to the same product and a near-identical description. I've deduped them and re-ordered the products from some version f best to worst.

Also bolded some of the specs to emphasize how to choose between the 3 options


sachaw:
Thanks for that, these changes will likely get integrated with some rewriting we're doing atm, so if I close this, consider it merged.

PR #254, opened 2022-02-22
Canned-messages doc update for release 1.3.

prampec:
The usage described in this change will be effective in version 1.3 and later.

PR #297, opened 2022-03-27
Canned-messages doc update for release 1.3.

prampec:
This is a rebuilt pull request of the existing #254, based on "plugin->module rename" changes.

Also, the issues noted in #254 are addressed here.


jfirwin:
Looks great. Thanks for making those changes!

PR #302, opened 2022-03-31
Update environment specific settings

thebentern:


jfirwin:
I have a PR that is dealing with the same stuff, plus the `inputbroker` material, so this PR will be closed by #303

PR #303, opened 2022-03-31
Fixing 1.3 module commands

jfirwin:
Fixing telemetry environment commands as well as renaming rotary encoder --> inputbroker


jfirwin:
closes #302

PR #306, opened 2022-04-01
Various Fixes

jfirwin:


jfirwin:
@mc-hamster I went out on a limb here and I'm showing how little I know about LoRa p2p. Could you review that section for me?

PR #319, opened 2022-04-22
Update range-test.mdx

bretbingham:
Has wrong 1.2 vs 1.3 "the name of this option has changed" ... just said "1.3 is range_test_ and 1.2 is range_test_" .... wasted 30 mins figuring this out -- should be 1.3 range_test_module_ and 1.2 is range_test_plugin_

### DIFF
--- a/docs/software/modules/range-test.mdx
+++ b/docs/software/modules/range-test.mdx
@@ -10,7 +10,7 @@ This module allows you to test the range of your Meshtastic nodes. It uses two n
 
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


jfirwin:
So sorry about that, thank you for the PR. I nabbed all the modules in the settings section and missed these ones in software apparently.

Nice catch!

bretbingham:
Least I could do to submit it when you guys are creating some awesome open
source goodness, keep up the great work you have a lot of fans out there!

PR #325, opened 2022-04-27
Change sidebar order

apt105:
A tidying up of sidebars, some filenames have been changed. Ideally every subfolder should have an index.html in future.


jfirwin:
Thanks for the PR! I'll work on the getting started section based off of your work.

PR #326, opened 2022-04-29
Update hardware pages and fix headings in downloads page

apt105:
Some further information added to the hardware pages for T-beam, T-echo and RAK. Also the CSS adjusted so headings now display correctly on the Downloads page. This CSS change doesn't appear to have affected the rest of the site.
https://meshtastic-git-small-fixes-apt105.vercel.app/docs/hardware/supported/tbeam


jfirwin:
Looking good! It's weird that the margins on the headings are non-existent. And that the `<a>` tags are not showing the link color until hovered over. I'm going to go ahead and merge this and we can assess those issues on a later PR.

apt105:
The margin situation will require a bit more tinkering. There is a margin: 0 arguments I found, but disabling it put margins around the top banner, so it won't be quite as straight forward.

PR #329, opened 2022-05-04
Update build.mdx

shivharis:
Minor changes that matter to the newly initiated.


jfirwin:
Thank you for the contribution! Hopefully one day I'll get in here and add tabs for each operating system, but this will help some users in the meantime.

PR #330, opened 2022-05-05
Update router.mdx

bretbingham:
Ls_secs default setting is listed as 5 minutes on this page, and 1 hour on power page in docs, so one must be wrong.

Here on router page, Under Details headiing, table line ls_secs, is this the correct “normal default” - 5 minutes?
On power page, The ls_secs default  power page isays the normal default is 1 hour


jfirwin:
Thanks for the contribution! According to [this](https://meshtastic.org/docs/software/other/power#events-that-decrease-cpu-activity) and [this](https://meshtastic.org/docs/developers/protobufs/api#radioconfiguserpreferences) the default is indeed one hour. Would you mind changing that in both places? Once that's done I'll be able to merge it in.

PR #334, opened 2022-05-07

bretbingham:
Using the recommended channel settings  "coding_Rate 4" causes critical error 7. When I looked at the coding_rate documentation on channels page it listed 5 as the lowest allowable setting.


jfirwin:
Good catch! Thanks for the contribution!

PR #336, opened 2022-05-11
fix/remove-inaccurate-note

andrew-moroz:
- removed note stating that exclamation point preceding node ID must be escaped


sachaw:
Great, if you could please just revert the change to the `.gitignore` as yarn is no longer used, thanks.

PR #338, opened 2022-05-13
iOS app doesn't share channel link yet

charminULTRA:
Edited to note that the iOS app doesn't have a share function yet.


jfirwin:
Good catch!

PR #339, opened 2022-05-14
Create flashing-external-serial-adapter.mdx

thebentern:
Here is my draft of the external usb to serial adapter document

### DIFF
--- a/docs/getting-started/flashing-firmware/flashing-external-serial-adapter.mdx
+++ b/docs/getting-started/flashing-firmware/flashing-external-serial-adapter.mdx
@@ -10,6 +10,10 @@ sidebar_position: 1
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
+:::info
+This information will likely only be helpful if you've already attempted to go through the prerequisites and processes outlined in [meshtastic flasher](https://meshtastic.org/docs/getting-started/flashing-firmware/meshtastic-flasher) or [manually flashing](https://meshtastic.org/docs/getting-started/flashing-firmware/flashing-esp32)
+:::
+
 :::caution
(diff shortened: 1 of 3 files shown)


docs/getting-started/flashing-firmware/flashing-external-serial-adapter.mdx
```diff
@@ -0,0 +1,78 @@
+---
+id: flashing-external-serial-adapter
+title: Flashing ESP32 Devices with an External Serial Adapter
+sidebar_label: Flashing with External Serial Adapter - ESP32 Devices
+sidebar_position: 2
```
jfirwin, line 5: Not sure what sidebar position we want for this one, position 1 probably makes sense, though we may need to bump the other `docs/getting-started/flashing-firmware/*` files numbers. I've been trying to use the `pagination_prev` & `pagination_next` to make the "Next" buttons walk the user through their workflow. I'm always pointing users to Meshtastic-flasher. So for this page I'd add the following:
```
pagination_prev: getting-started/flashing-firmware/flashing-firmware
pagination_next: getting-started/flashing-firmware/meshtastic-flasher
```

I think `docs/getting-started/flashing-firmware/index.mdx` should bypass this page as this is kind of an "advanced" use case. So on index.mdx we can add:
```
pagination_next: getting-started/flashing-firmware/meshtastic-flasher
```
and on `meshtastic-flasher.mdx` we can add
```
pagination_prev: getting-started/flashing-firmware/flashing-firmware
```

jfirwin (commented):
This looks great! All I really think we need is some additional yaml frontmatter on a couple of files.

Only other thing I could think of is an admonition letting users know to skip the page if they've successfully gotten past the driver/data cable section in the prerequisites. That'll only effect users who aren't clicking the next buttons though.

PR #341, opened 2022-05-15
Cleaned up documentation, wrote some pros/cons

charminULTRA:
I edited the section to hopefully make it easier to read. I'd also propose we add a warning that the T-Echo is very buggy currently and people should be warned to check the current status of bugs or check with the community before buying, but I'll leave this up to some else's discretion.

### DIFF
--- a/docs/hardware/index.mdx
+++ b/docs/hardware/index.mdx
@@ -43,7 +43,8 @@ The following nRF52 devices are currently supported:
 
 Advantages:
 - Smaller, more compact designs are available
-- Very low power usage due to Bluetooth 5.0 (BLE)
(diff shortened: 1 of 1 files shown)


garthvh:
Do you mind adding a recommendation that that NRF52 have 10-20x less battery draw?  That is the other relevant spec.

charminULTRA:
> Do you mind adding a recommendation that that NRF52 have 10-20x less battery draw? That is the other relevant spec.

Updated to say it draws 90-95% less power, does that work?

jfirwin:
I think that works well unless @garthvh has any objections? If not, we can go ahead and merge this.

garthvh:
Looks good

PR #342, opened 2022-05-16
Removed mention of /static/

charminULTRA:
Removed mention of /static/ as the Range Test output is no longer located in this directory.


jfirwin:
Thanks for the PR! I have a couple more edits to make, but this (along with several other locations `static` is mentioned will be covered in my PR #337. You also previously mentioned there's nothing tying the software/modules and settings/modules together, so that's part of what I'm workin on on that PR too.

Closing this, and I'll get mine merged in today. Thanks!

PR #345, opened 2022-05-26
./ is not used in Windows shell

charminULTRA:
Using ./ at the start of the shell command in Windows 10 will result in an error, so I've removed this part of the instruction.


jfirwin:
Originally I wrote these instructions using Gitbash. There may be other hangups using windows shell. As it stands, are all the commands working minus this change or should we add some extra edits to this PR?

jfirwin:
Perfect! Just what I wanted to verify!

PR #349, opened 2022-05-27
Update meshtastic-flasher.mdx

charminULTRA:
Updated to reflect the information inside https://github.com/meshtastic/Meshtastic-gui-installer which states that the single install executables were deprecated in March 2022. This page was previously linking to versions of the Flasher from March 2022 (approx 84 commits behind).

### DIFF
--- a/docs/getting-started/flashing-firmware/meshtastic-flasher.mdx
+++ b/docs/getting-started/flashing-firmware/meshtastic-flasher.mdx
@@ -99,74 +99,11 @@ sudo apt install -y python3 python3-pip python3-venv
 
 ### Install App
 
-This is the preferred installation method for `meshtastic-flasher`.
 
-<Tabs
-groupId="operating-system"
-defaultValue="linux"
-values={[
-{label: 'Linux', value: 'linux'},
-{label: 'macOS', value: 'macos'},
-{label: 'Windows', value: 'windows'},
-]}>
-<TabItem value="linux">
-
-```shell title="Install Meshtastic Flasher"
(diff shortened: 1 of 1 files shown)


charminULTRA:
Just noticed the Winapp auto-updates to the latest release, so maybe this PR isn't necessary? That said, might make sense to remove the deprecated note in https://github.com/meshtastic/Meshtastic-gui-installer so that there isn't conflicting documentation?

jfirwin:
I've added someredguy/aka @mkinney to review this. He'll know for certain which way is right but I'm leaning towards our link on the download page should reflect what is currently on production on this page. (Including the use of tabs, it may be a nice touch over there)

charminULTRA:
> I've added someredguy/aka @mkinney to review this. He'll know for certain which way is right but I'm leaning towards our link on the download page should reflect what is currently on production on this page. (Including the use of tabs, it may be a nice touch over there)

The only thing about linking the Download page is that, that page points to the most recent release, which is just source code.

jfirwin:
Yeah sorry, I was saying basically the tabs this PR deletes should be replace the button on the downloads page. The Linux installation instruction is a little verbose for the downloads page so I would work that to link back here.

jfirwin:
Sorry @charminULTRA, life hit hard the past couple weeks. I haven't been able to do much open source work. If I recall correctly from discord, we want to revert the changes made here and add the Tabs functionality to the downloads page. Should I close this PR?

garthvh:
This is correct

PR #350, opened 2022-06-01
Proposal to delete page

charminULTRA:
This page looks to be a collection of to-do's and notes, which I propose should be moved to a projecting space, rather than user-facing documentation.


jfirwin:
I'm in favor of this unless there's any objection. Just want to make sure the notes aren't totally lost. @caveman99 is there any info here you or any other devs would like to keep somewhere?

caveman99:
I think this was moved from the software tree to doc at some time. The notes are partly valuable so we should keep them somewhere. It's certainly relevant to development, even if only in retrospect. I remember the 'hunt for the elusive T-Echo Bootloader' ;-) Can we create a section ' Developer Notes' on the 'Developer' docs section?

jfirwin:
Let's move the file to a developers notes section like caveman suggested. Then we can merge this PR in.

garthvh:
Going to merge this shows up in google and is confusing, git has a history if we need these precious notes again

PR #354, opened 2022-06-24
fix region enum

romanimm:


caveman99:
The docs are transitioning from 1.2 to 1.3 firmware and there the EU region is indeed named EU868 with a different center frequency. this will be adjusted accordingly. EU865 is the old 1.2 name which will soon be obsolete.

PR #356, opened 2022-06-25
Update according to device PR #1500

GUVWAF:
I updated the mesh algorithm according to PR #1500 of meshtastic-device. Furthermore, I added a link to the simulator for the people that want to dive deep into it (but if you don't want this on the website I can remove it).


garthvh:
Awesome work

PR #400, opened 2022-09-07
1.2 End of Life

garthvh:
Move all docs under 1.2 End of life menu item


garthvh:
I moved everything 1.2 instead of deleting it since we will still need some of it to get the 1.2 docs into shape

garthvh:
Python API has not been updated in PIP so 1.3 is the latest version and not a pre-release anymore

garthvh:
Same for the flasher

PR #401, opened 2022-09-11
param is case sensitive

osmanovv:
Using Python CLI v1.3.29:
```
>meshtastic --set device.role client
Connected to radio
device.role does not have an enum called client, so you can not set it.
Choices in sorted order are:
    Client
    ClientMute
    Router
    RouterClient
LocalConfig and LocalModuleConfig do not have attribute device.role.
```


garthvh (approved):
Looks good

PR #402, opened 2022-09-11
param value is case sensitive

osmanovv:
Using Python CLI v1.3.29:
```
>meshtastic --set bluetooth.mode randomPin
Connected to radio
bluetooth.mode does not have an enum called randomPin, so you can not set it.
Choices in sorted order are:
    FixedPin
    NoPin
    RandomPin
LocalConfig and LocalModuleConfig do not have attribute bluetooth.mode.
```


garthvh (approved):
Looks good

PR #404, opened 2022-09-11
param value is case sensitive

osmanovv:
Using Python CLI v1.3.29:
```
>meshtastic --set lora.region eu433
Connected to radio
lora.region does not have an enum called ru, so you can not set it.
Choices in sorted order are:
    ANZ
    CN
    EU433
    EU868
    IN
    JP
    KR
    NZ865
    RU
    TH
    TW
    US
    Unset
LocalConfig and LocalModuleConfig do not have attribute lora.region.
```


garthvh (changes requested):
UNSET should be all caps too

garthvh (approved):

PR #406, opened 2022-09-11
Added missing `Very Long Range - Slow` channel and fix default to `Long Fast`

osmanovv:
Needs to check data rate and other channel params


garthvh (changes requested):
Please update VLongSlow to Very Long Slow as well

garthvh (approved):

PR #426, opened 2022-10-03
Update rak4631.mdx

thedeepself:
Removed baseboard model 5005/19003 from firmware filename and changed epaper firmware filename to eink.


andrekir (approved):
thanks!

PR #431, opened 2022-10-05
Update encryption.mdx

nymble:
Added a proposal for phased improvements in security framework starting with 'strong identity' for clients and devices based on public key techniques.


docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+### Phase 1 - Fixed network encryption with AES-CTR
+
+The current implementation provides optional confidentiality to members of a configured network:
+ * Encryption is implemented in devices/nodes with network-wide encryption keys.
+ * Encryption is optional and is turned off when devices are in 'Ham mode'.
+ * There is no encryption supported in the clients (iOS, Android) to facilitate distribution as mass market software.
+ * Pairing from client-to-device is by:
+    * direct USB cable
+    * BT pairing
+ * Devices are 'promiscuous' and will pair with any near-by client. Network confidentiality requires physical protextion of all nodes.
```
sachaw, line 68: Not sure if I like the wording of confidentiality, I think there should be a distinction between if a node gains anonymity or security through encryption

Also, physical "protection" will by no means prevent unauthorized use of a node (bar serial access)

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+
+### Phase 2 - Strong device and client identity
+
+**Phase 2 security goals:**
+ * Know who sent a message (strong binding of messages to a particular node and/or terminal device)
+    * This would be an optional feature for a message
+ * Optionaly enforce identity based restrictions on some actions performed at nodes and/or clients
```
sachaw, line 75: as per my last point, the entire payload could be signed

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+ * Optional support of strong pairing of a client to a device/node and restrict ability to manage and receive messages based on the pairing.
```
sachaw, line 76: by "cryptographic pairing", are you referring to key exchange?

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+ * Above features should be architected to be ‘cryptographically strong’ and algorithm agile.
+
+**Phase 2 Proposed mechanisms:**
+ * Proposed initial algorithms
+    - Ed25519 for signatures based on NaCl libraries and iOS support for Ed25519
+ * Clients and nodes to generate local identity Ed25519 keys 
+ * Devices maintain knowledge of owner public key
+ * Devices maintain knowledge of some peers public keys and associated information (name, etc.)
+ * Experimental protobuf message type with
+     * cipher suite indicator (csi)
```
sachaw, line 87: this should be a flagged enum field

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+     * wrapped data using a cipher suite identifier to indicate use of Ed25519 wrapping identified algorithms.
```
sachaw, line 88: not sure what you're getting at here, if you mean for this to be the generic encrypted payload, we already have a place for that

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+ * Proposed initial algorithms
+    - Ed25519 for signatures based on NaCl libraries and iOS support for Ed25519
+ * Clients and nodes to generate local identity Ed25519 keys 
+ * Devices maintain knowledge of owner public key
+ * Devices maintain knowledge of some peers public keys and associated information (name, etc.)
+ * Experimental protobuf message type with
+     * cipher suite indicator (csi)
+     * wrapped data using a cipher suite identifier to indicate use of Ed25519 wrapping identified algorithms. 
+       Wrapped data to contain any of the existing message types.
+     * initial ‘cipher suite’ **only** signs a message
```
sachaw, line 90: I think the concept of a ‘cipher suite’ should be limited to alg selection and not cover anything to do with signing

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+     * new signed/authenticated messages to:
+      *  device->client:  provide ownership status of device (owner is identifed by a public key)
+      *  client->device:    set owner key (must be existing device owner or owner null)
+      *  any->all.  Broadcast public key and associated info (crude initial key distribution)
```
sachaw, line 94: I think public key broadcast is fine in all cases
by other info, I assume you mean the cipher suite. I think that should happen on request (first direct message after initial public key exchange has happened to prevent network congestion)
And possibly make the public key be broadcast on request

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+ 
+### Phase 3 and 3+:
+ * Device/node e2e message confidentiality (1-to-1)
```
sachaw, line 97: I think we need to drop the confidentiality part all together, and just specify if the payload is encrypted or not, and by whom channel key (device side) or private key (client side)

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+    * NO client side encryption
```
sachaw, line 98: I think this could become an option for super encryption packets, imaging you want the public key advertisements to be encrypted with the channel key and then the private key signed traffic to occur still within that channel (more overhead but an option)

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+    * All encryption turned off for a network in 'Ham mode'
```
sachaw, line 99: We still need a far better definition of what ham mode means for our devices, verily vague atm

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+  * Protect messages sent over LoRa from eavesdroppers outside of a well identified group
+       - multicast group key distribution / discovery (1-to-n)
```
sachaw, line 101: 'multicast groups' would be what we already conceptualize as channels

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+ * Privacy 
+    * BT MAC layer address privatization
+    * Node address privatization (Use use if based on device public key and schedule based aliasing)
```
sachaw, line 104: If you want to push the privacy aspect, there's a possibility of having ephemeral node id's, but would require a fair bit of work.

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+    * Private peer discovery
```
sachaw, line 105: I'd be interested in how you would facilitate this, I don't see much of a benefit really

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+ * Private service announcement and discovery
```
sachaw, line 106: could you please elaborate?

sachaw (commented):
Great work, I'd love to progress this so a more fleshed out specification that we can work towards.

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+
+### Phase 1 - Fixed network encryption with AES-CTR
+
+The current implementation provides optional confidentiality to members of a configured network:
+ * Encryption is implemented in devices/nodes with network-wide encryption keys.
+ * Encryption is optional and is turned off when devices are in 'Ham mode'.
+ * There is no encryption supported in the clients (iOS, Android) to facilitate distribution as mass market software.
+ * Pairing from client-to-device is by:
+    * direct USB cable
+    * BT pairing
```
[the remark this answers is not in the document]
nymble: Oh - did not try that yet, will add.  
The ideas I'm documenting originally come from Wi-Fi pairing work.

docs/developers/Firmware/encryption.mdx
```diff
@@ -52,3 +52,63 @@ I'm assuming that meshtastic is being used to hike in places where someone capab
...
+ * Devices are 'promiscuous' and will pair with any near-by client. Network confidentiality requires physical protextion of all nodes.
+
+### Phase 2 - Strong device and client identity
+
+**Phase 2 security goals:**
+ * Know who sent a message (strong binding of messages to a particular node and/or terminal device)
```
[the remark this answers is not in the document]
nymble: This would be 'data origin authentication' and 'data integrity'
Yes - message sizes will be reduced ... by 65 bytes, 64 for signature, one byte for cipher suite identifier to support later algorithm agility ... and perhaps other later optimizations (use of a keyed MAC instead of a public key signature)

docs/developers/Firmware/encryption.mdx
nymble: Not necessarily. Ideally we would use a privacy preserving key exchange (like SIGMA-I) ... but that introduces encryption on the client. 
I'm using pairing here to describe that client and device have determined each others public key in a trusted manner. This always requires some type of out-of-band trusted path. The simplest is a USB cable. Having the client sign a nonce over BT pairing with option checking of a few digits of an ID is pretty strong.
Long term ... a un-configured device could display a QR code that would be use by the client to support simple pairing.

docs/developers/Firmware/encryption.mdx
nymble: ... this is not encryption ... trying to support client-to-device trust.  Might be useful to get separate for simplicity of export control concerns.
This - could just be a "protected" message, with cipher suite byte indicating AES-CTR encryption or Ed25519 signed
I feel initially a separate new "protected" generic type from the current encryption type would ease integration. Would break less as this is implemented. Having the new strong identity work be in new messages is safer near term.

docs/developers/Firmware/encryption.mdx
nymble: ... the first one or more bytes of a protected message (encrypted, signed, whatever) should provide everything necessary to apply the cryptographic processing. Long term this should be designed to enable MAC based integrity, key ratcheting etc.  
On a slightly related topic ... on how to determine the cryptographic process...
The cryptographic processing should use other header fields. The node addresses can be used to determine the keys/session to use. Encrypted messages should be using an AEAD to protect the header fields.
I believe the 'only sign' for now might mitigate concerns about describing this work for export review.
This discussion also overlaps with the 'generic encrypt' versus 'protect' versus 'signed' message types.
Similar concepts were debated in IPSec for the ESP and AH modes of operation.  AH was created to provide integrity only (MAC based, not signed). It was barely used and could have been handled by ESP cipher negotiation. In the case of the Meshtastic environment. It would be useful to clearly be able to identify encrypted messages versus signed messages for 'Ham mode'

docs/developers/Firmware/encryption.mdx
nymble: Associated info could include: short name, node identifier
This would be a simple signed certificate ...
Yes - should be able to request transmission of this 'certificate' from peers

docs/developers/Firmware/encryption.mdx
nymble: Section 4B of: https://www.recercat.cat/bitstream/handle/2072/355151/alwayson-2015.pdf

docs/developers/Firmware/encryption.mdx
nymble: Some people value privacy. Current design uses constant MAC addresses and node identifiers.
Private peer discovery could be a SIGMA-I type exchange (e.g. TLS 1.3) or HPKE (RFC 9180) encapsulating the long term certificate 
Node address privatization is what Apple is doing with MAC address rotation and also the find-my public key usage.

docs/developers/Firmware/encryption.mdx
nymble: "Well identified" meaning using the node public keys to authenticate and distribute keys. This would provide remote/dynamic updates in keys and group membership.

docs/developers/Firmware/encryption.mdx
nymble: Yes. 
I'm assuming that a 'Ham mode' device should never relay any encrypted messages.  With stronger 1-to-1 encryption, the messages need to be identifiable so they can be dropped by a 'Ham mode' device.

docs/developers/Firmware/encryption.mdx
nymble: Good point on the signed and public key announcements being within the current encryption messages.
I need to study the formats more closely ...
On client-side, strong pairing of client-to-device could provide e2e protection even without client-side encryption. The locally controlled node would be acting on behalf of the client for e2e encryption. This could be supported by client side signed messages linking the local node (acting as a crypto coprocessor of sorts). 
Effectively - the a client would give the node public key as proxy for it's end address.  
Longer term discussion of key lifetimes, changing node-client pairing etc.   But as a long term goal I'd like to see: " nodes acting as client proxy for e2e message confidentiality"

docs/developers/Firmware/encryption.mdx
nymble: Confidentiality is a 'service' and makes sense as a requirement. Encryption is a mechanism. In this context they are interchangeable....

docs/developers/Firmware/encryption.mdx
nymble: Yes. Makes sense to have devices to request keys from peers. 
Was also trying to make this phase a very minimal PoC.

docs/developers/Firmware/encryption.mdx
nymble: Also - multicast is a subset of the connected network that can be reached by broadcast.

docs/developers/Firmware/encryption.mdx
nymble: Currently Meshtastic only supports simple delivery of information and some device related data. There is no layered model for link/network/application.  Services in this context should be considered to be like TCP ports ... a layer 4-to-7 end-point to participate in particular arranged set of communications.  Ports are static identifiers and leak privacy info. 
A 'service' for Meshatastic might something like ... a device announcing the ability to forward to email.
Private service discovery is potentially just a bloom filter announcement from a node. 
But this was a long term idea for discussion ...

sachaw:
Some more conversation is definitely gonna come from this, I'm going to merge this for now whilst we focus on 2.0, once that's over, we can move this up in priority

sachaw (approved):

PR #447, opened 2022-10-22
Standardize LoRa region codes

### LINKED ISSUE #446: LoRA region codes do not match values in Python CLI
https://github.com/meshtastic/Meshtastic/blob/26bafe5a81fa2c406329849b76106f8a77451bd4/docs/settings/config/lora.mdx?plain=1#L21

The Python CLI documentation points to this page to set preference fields, however the region codes in the table do not match the values accepted by the Python CLI.

Update this table to reflect the acceptable values when setting the LoRa region in the Python CLI.

ahmedkadd:
Close #446 

Update the LoRa region codes in the table to match the values accepted by the Python CLI.


garthvh (approved):
Looks good

PR #454, opened 2022-10-26
Update remote-admin.mdx

charlesmarseille:
specify ID format for windows CLI


garthvh (changes requested):
Can you remove the parens, those are going to get copied by users as is.

garthvh (approved):

PR #494, opened 2022-11-03
Update installation.mdx

Dtrieb:
Add WLAN Label.


caveman99 (changes requested):
That doesn't seem to fit thematically. I'd rather remove that obscure WLAN Comment.

garthvh:
> 

Same

sachaw:
Not sure why this was added without any content.

PR #497, opened 2022-11-07
Update faq.mdx

Dtrieb:
This part was doubled :

Why does only one RAK Meshtastic Starter kit show up in my node list? There was a bug where Meshtastic Starter kits were sent out with the same MAC address. With a single MAC address the devices all report as being the same device. Without the battery connected, flash the starter kit device(s) to 1.2.59 and then do a factory reset, disconnect and reconnect the board and run meshtastic --info.

https://meshtastic.org/docs/faq#why-does-only-one-rak-meshtastic-starter-kit-show-up-in-my-node-list https://meshtastic.org/docs/faq#why-does-only-one-rak-meshtastic-starter-kit-show-up-in-my-node-list-1


garthvh (changes requested):
Can you update this to say any firmware > 1.2.59

garthvh (approved):
Thanks!

PR #498, opened 2022-11-08
Credits

jfirwin:
My work in progress `/credits` page. A preview can be seen [here](https://meshtastic-git-credits-jfirwin.vercel.app/credits).

Do we want the API calls to Open Collective & GitHub to occur in real time when a person loads the page? Or should we batch things up and put it together when we build the project like the protobufs are managed? To get the GitHub contributors for an organization we need to grab each repo, grab the contributors, and combine it all. I think it's a lot of calls to be happening every time a user opens the credit page, so that's why I'm leaning towards batching it up with each build. @sachaw, is there a way we can store the resulting json object on vercel at build? I haven't done too much populating templates like the protobufs are set up.


sachaw:
I'd say realtime

caveman99:
i'd collect it on build. getting all that stuff from github will trigger the anaonymous API limit quite fast, if the client browser does it.

sachaw:
Or have an endpoint on our API and cache it

sachaw:
I'll merge this for now, I'll add endpoints on our api to cache the required info that we can fetch.

sachaw (approved):

jfirwin:
Awesome. We can move the avatar grid and avatar components into utils, and add extra optional items into the interface to expand it for say, the admin bios section.

Also some of the copy about the donations I might condense once we have additional content on the page.

PR #500, opened 2022-11-08
Downloads page edits

jfirwin:
made some minor tweaks to the download page. Used the `card` elements that we've used elsewhere on the page to give it a more consistent look and feel. CSS looks pretty good and responsive, just the clients.meshtastic.org external link icon sometimes disappears. I'll try to take a look at that tomorrow.


jfirwin:
Edits viewable [here](https://meshtastic-git-downloads-page-edits-jfirwin.vercel.app/downloads)

sachaw:
I'll eventually make a card component what we can use bug good for now!

PR #504, opened 2022-11-09
Update installation.mdx

rcarteraz:
Added a section for using obtainium to install/update directly from the Meshtastic GitHub releases page.


sachaw:
good addition, thanks.

sachaw (approved):

andrekir:
sorry, I can only support the sources we currently offer.

andrekir:
@caveman99 that is true for an endless list of alt-sources that grab apks from the play store, f-droid, github. I can't check what every alt-source app does under the hood or the result install apk, and can only vouch for the sources I manage.

PR #507, opened 2022-11-14
Update mqtt.mdx

samuelhj:
Added mentioning of uplink and downlink on channels, in case someone else doesn't notice that part while configuring mqtt.

### DIFF
--- a/docs/configuration/module-config/mqtt.mdx
+++ b/docs/configuration/module-config/mqtt.mdx
@@ -8,7 +8,7 @@ sidebar_label: MQTT
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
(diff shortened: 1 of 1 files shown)


docs/configuration/module-config/mqtt.mdx
```diff
@@ -8,7 +8,7 @@ sidebar_label: MQTT
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-If your device is connected to WiFi you can enable it to forward packets along to an MQTT server. This allows users on the local mesh to communicate with users on the internet.
+If your device is connected to WiFi you can enable it to forward packets along to an MQTT server. This allows users on the local mesh to communicate with users on the internet. At least one channel must be enabled as uplink and/or downlink for the MQTT module to work. See [channels](/docs/settings/config/channels#downlink-enabled)
```
caveman99, line 11: Strictly speaking this is not correct. The node will still send status messages, like 'i just rebooted and connected' just no protobuf content.

caveman99 (changes requested):

docs/configuration/module-config/mqtt.mdx
caveman99: also MQTT is enabled for wired ethernet too... :-)

samuelhj: I see, perhaps we can paraphrase what I wrote to reflect that? This is something I hadn't realised either when I was testing.

caveman99 (approved):

PR #525, opened 2022-11-26
Added a node-red example on encoding Position protobuf off-device

arduionoGP:


caveman99 (approved):
No idea how this will look in docosaurus. Will it pretty print the JSON? Can be changed after merge fi it doesn't look good :-)

arduionoGP:
Hopefully, the flow code will come out as a single, very long line that can be cut and pasted for import into node-red. If it is more than a single line, will need to change it to a file link.

PR #528, opened 2022-11-29
update hardware notes

caveman99:
add Heltec V3 and deprecation note for older boards add clarification on RAK boards


garthvh (approved):
Looks good

PR #530, opened 2022-12-01
clarify serial module operation

caveman99:


jfirwin (approved):
Everything appears to be correct here. Though I personally haven't tested the serial module.

PR #532, opened 2022-12-04
Update index.mdx

arduionoGP:
Update docs for the ability to send JSON Position to device assuming something like pull request #2018 for firmware gets included in 2.0.7.


caveman99 (dismissed):
The next release will have that included. Can you please change the wording as if it was already there? Changing texts every few days related to versioning is not very effective.

garthvh (approved):

PR #535, opened 2022-12-11
add RAK GPIO mapping

caveman99:


garthvh (commented):
Can we add a note that there is no usable GPIO pins on any base board other than the 19001 without adding an IO module. We keep having to re-explain this point over and over.

garthvh (approved):
Looks great!

PR #538, opened 2022-12-13
css fix for social icons on lg and smaller screens

jfirwin:
Check this on iPad down to iPhone SE and the social icons are now scrolling instead of making the whole page scroll. Viewable [here](https://meshtastic-qbb8fmpd7-jfirwin.vercel.app).


garthvh (approved):
Looks good

PR #552, opened 2022-12-30
Mermaid diagram theme

jfirwin:
Added Mermaid Diagram support. Color Pallet based off of this:
Screenshot 2022-12-30 at 1 50 21 PM

`--tw-prose-headings` was used rather than `--tw-prose-body` because it offered greater contrast for readability. This allowed the text to adhere to the users light/dark selection on the website. I had to install mermaid, so there are some changes to the `package.json` and lock file. If we need to remove those changes I'm happy to do that.

No diagrams have been implemented yet, but they will be themed to our color scheme when they are.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -122,6 +122,10 @@ const config = {
   customFields: {
     API_URL: process.env.API_URL,
   },
+  markdown: {
+    mermaid: true,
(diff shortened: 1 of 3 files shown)


sachaw (approved):
Exceptional work!

PR #574, opened 2023-01-21
Updates to Getting Started

rcarteraz:
I'm taking a stab at #438 

I have made several changes to this section to try and make it more user-friendly and improve the flow from one step to the next. While this isn't a comprehensive overhaul, it should help to make things clearer for users in the meantime.

Here are the changes I made:

- I condensed the main "Getting Started" index page. It was a bit cluttered trying to be both a one-page guide and a direct to separate pages for each step. Now, it's more like a table of contents that directs you to the specific step you need and utilizes the existing pages we have rather than trying to cram everything into one page.
- I created a separate page for "Test Serial Driver Installation" and linked it in both the "ESP32 Drivers" and "nRF52 Drivers" sections. This way, users only need to visit this page if they're unsure if their driver is working (or want to test it to be sure) and can then proceed to flashing firmware.
- I added button links to each page that direct users to the next step in the process. Additionally, I removed the "target="_blank" from these links so that when clicked, they open in the current tab instead of a new tab. This helps to prevent confusion about what step of the process the user is on and where they should go next. For example, at the bottom of the "ESP32 Drivers" page, there's a button link to "Flash ESP32 Firmware." At the bottom of each method's page, there's a link for "Connect and Configure Device," and at the bottom of the "Initial Configuration" page, there's a link for "Device Configuration." Same with nRF52 pages.
- I made a minor edit to the css as for some reason I could not get some of the button links to split left and right like they were configured to, even though others were. I'm not sure why but adding this little bit and changing the class worked. Happy to revert this if someone wants to fix that issue. It's cleaner when they're all split left and right.

Please review the vercel preview of these changes. Navigate through the different pages to test the flow, and let me know if there's any changes that should be made. 

[Preview Link](https://meshtastic-docs-qkkdhnd3j-rcarteraz.vercel.app)

### DIFF
--- a/docs/getting-started/flashing-firmware/esp32/cli-script.mdx
+++ b/docs/getting-started/flashing-firmware/esp32/cli-script.mdx
@@ -1,12 +1,13 @@
 ---
 id: cli-script
 title: Flashing with the CLI
-sidebar_label: CLI Script
+sidebar_label: CLI Script (Advanced Users)
 sidebar_position: 3
 ---
 
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
+import Link from "@docusaurus/Link";
 
 :::caution
 Make sure not to power the radio on without first attaching the antenna! You could damage the radio chip!
@@ -269,8 +270,15 @@ Be very careful to install the correct load for your board. In particular the po
 
 Select `Flash ESP`. It may take a minute or two. Once complete, "Done! Flashing is complete!" will be shown.
 
-## Over the Air Update Instructions
+## Connect and Configure Device
 
-OTA updates are only currently available on Android.
+After flashing the Meshtastic firmware to the device, you can proceed with the initial configuration.
 
-<!--- TODO --->
+<div className="indexCtasBody">
+  <Link
+    className={"button button--outline button--lg cta--button"}
+    to={"/docs/getting-started/initial-config"}
+  >
+    Connect and Configure Device
+  </Link>
+</div>
--- a/docs/getting-started/flashing-firmware/esp32/index.mdx
+++ b/docs/getting-started/flashing-firmware/esp32/index.mdx
@@ -11,7 +11,7 @@ The recommended method for firmware flashing is the [Web-Based Installer.](https
 
 ## Flashing Method for ESP32 Devices
 
-1. The [Web-Based Installer](https://flasher.meshtastic.org) requires either Chrome or Edge browsers but is an excellent choice for quickly flashing devices. **This is the recommended method for firmware flashing, especially for those new to the project, due to its ease of use.**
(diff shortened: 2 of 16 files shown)


docs/getting-started/flashing-firmware/esp32/index.mdx
```diff
@@ -11,7 +11,7 @@ The recommended method for firmware flashing is the [Web-Based Installer.](https
 
 ## Flashing Method for ESP32 Devices
 
-1. The [Web-Based Installer](https://flasher.meshtastic.org) requires either Chrome or Edge browsers but is an excellent choice for quickly flashing devices. **This is the recommended method for firmware flashing, especially for those new to the project, due to its ease of use.**
+1. The [Web-Based Installer](/docs/getting-started/flashing-firmware/esp32/web-flasher.mdx) requires either Chrome or Edge browsers but is an excellent choice for quickly flashing devices. **This method is highly recommended for firmware flashing, especially for new users of the project, as it is easy to use.**
```
jfirwin, line 14: Two things.

1. I'm surprised to se the web flasher instructions nested in the ESP32 pages (it'll work for both, though I know we want to push for drag and drop on nRF52)
2. Should we keep the link going to flasher.meshtastic.org and instead have a button added to that the flasher page back to the docs instructions?

jfirwin: I must be confused on point 1, I thought it was capable of flashing the nrf52 devices as well!

I just took a look at the vercel preview you provided. It looks great. I have no issues with of the changes. I think you've added some additional clarity and the cta buttons are great to help the user click through the flow.

jfirwin (approved):

docs/getting-started/flashing-firmware/esp32/cli-script.mdx
```diff
@@ -269,8 +270,15 @@ Be very careful to install the correct load for your board. In particular the po
...
 Select `Flash ESP`. It may take a minute or two. Once complete, "Done! Flashing is complete!" will be shown.
 
-## Over the Air Update Instructions
+## Connect and Configure Device
 
-OTA updates are only currently available on Android.
+After flashing the Meshtastic firmware to the device, you can proceed with the initial configuration.
 
-<!--- TODO --->
+<div className="indexCtasBody">
```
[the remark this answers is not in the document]
jfirwin: I think in the getting started section we have enough diverging in the flow, `null` could be a good option. Let's do that on a future PR.

Gary-in-Geneva:
Please update the instructions for the ESP flashing.

For flashing the ESP32-S3 (Heltec HTIT - Tracker) using the web based installer the device needs to be put into flash mode.

You enter flash mode as follows:
Pressing and holding the reset button
Press and hold the user button
Release the reset button
Release the user button. 

After doing that the board flashes 100% of the time.

PR #583, opened 2023-01-30
Added modem preset Long Moderate

charminULTRA:


charminULTRA:
Oops, messed something up in a prior PR

PR #585, opened 2023-01-30
Update Device Configuration

rcarteraz:
@charminULTRA and I collaborated on an update the Device Configuration page to include both of the new roles (`REPEATER` and `TRACKER`) as well as changing the language on some of existing roles. The language was updated to be more plain English to help clarify for users what the role would be used for. 

We also added a new section detailing "Rebroadcast Mode" and the options available with the `device.rebroadcast_mode` setting. 

[Vercel Preview](https://meshtastic-docs-qiwf0wul1-rcarteraz.vercel.app/docs/settings/config/device)


GUVWAF:
PR [2232](https://github.com/meshtastic/firmware/pull/2232) that is just in also prefers packets to be routed over a repeater, in the same way as for a Router. 

Furthermore, to make the benefit of a repeater more clear, I would add to this:

> This role is useful for a cleaner mesh setup where users do not need information about the Repeater node

that it also eliminates all the overhead that a normal node would add to the network (sending NodeInfo, DeviceTelemetry, etc.).

GUVWAF:
> or delete the rebroadcast bit?

Yes, I think you can delete the rebroadcast bit and add that because they are not showing up as part of the network, there is less overhead.

GUVWAF:
@rcarteraz Perfect!

thebentern (approved):

PR #595, opened 2023-02-17
Update atak.mdx

jeremybox:
Google Drive link has been deprecated.  Updated document to reflect the link to github, and remove references to Google Drive to prevent confusion.


caveman99:
Could you add an info that this is only signed for the downloadable version from tak.gov and not the play store version?

caveman99 (approved):

PR #601, opened 2023-03-28
include Neil's ADC calc with attribution

caveman99:


thebentern (approved):

caveman99:
any idea why the tests are failing all of a sudden with weird complaints? was working yesterday ;-)
Or do we just need to add that flag to the build command?

caveman99:
looks like https://github.com/pnpm/action-setup/issues/40 is hitting us.

PR #650, opened 2023-06-02
make tab items buttons

jfirwin:
users have suggested that the tabs supplied from docusaurus don't look interactive enough, so we've made them look more like buttons and I added some additional color on the :hover attribute.

Vercel link [here](https://meshtastic-i6dtfb3rz-jfirwin.vercel.app/docs/settings/config/bluetooth)


jfirwin:
Thanks! I'm happy to toy with a cleaner look if we want go that route. I saw similar buttons on Speedtest.net and thought it might be an ok compromise too.

jfirwin:
Sorry on the delay here, I ran into some weird inheritance issues with the active class and didn't finish it before getting pulled away. I'll try to dig in this weekend.

tekstrand:
Hey I wasn't aware of this until now - I have many ideas! Our core issue, IMO, is 2 fold.
1. An unclear inactive state
2. A lack of visual connection between the tab controls and the tab content

I would suggest we take one or both of these approaches
1. Underline the inactive tabs with our inactive state color https://codepen.io/flkt-crnpio/pen/WxROwy/
2. Create a background for the tabs and the content within them, connecting them both to each other, and their content https://codepen.io/alvarotrigo/pen/GRMbzBR

tekstrand:
Yes! Last thing that we are missing, if we don't want to change the background color of the content, is to make it the first tab line up on the left the same as the content. Right now the tabs are floating in space.

PR #693, opened 2023-06-29
Remove flasher from downloads page

pdxlocations:
Flasher is no longer download, so remove it from downloads page.
[preview link](https://sandbox-git-flasher-not-downloads-pdxlocations.vercel.app/downloads)


pdxlocations:
Yes, my point here is that these links on the downloads page are not links to downloads.  They are links to a web app and page of instructions.  They don't make sense on this page.  Flashing used to be a download but it isn't anymore.

They are good links to feature so perhaps leaving them here makes sense but I think they are promoted enough times elsewhere.  I certainly won't be offended if this PR isn't merged.

pdxlocations:
> It's a single page resource that directs people in the right direction.

Yeah, totally!

PR #717, opened 2023-07-08
Update trademark-grants.mdx

markbirss:
Grant: Mark Birss the use of "Meshtastic" , Meshtastic.org URL and "M" logos.


garthvh (approved):
Awesome!

PR #728, opened 2023-07-15
Update serial.mdx

htcheroportugal:
Add Interfacing relay module With External Microcontroller


htcheroportugal (commented):
add name of code author

PR #738, opened 2023-07-19
correction of diagram

htcheroportugal:
please put the image on the right link, because i used imgbb


htcheroportugal (commented):
please use the correct link to use the image

htcheroportugal:
@rcarteraz thanks for your fix , i am still learning. but i hope now everthing its fine

Issue #742, opened 2023-07-23
Incorrect or out of date MQTT topic description?


tobymurray:
[Docs say](https://meshtastic.org/docs/software/mqtt/#topics):
> The "mesh/crypt/CHANNELID/NODEID/PORTID" [topic](https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices) will be used for messages sent from/to a mesh.

I'm seeing topics that look like: "msh/2/c/LongFast/!abf849b0", meaning:

mesh = "mesh"
crypt = "2"
CHANNELID = "c"
NODEID = "LongFast"
PORTID = "!abf849b0"

This doesn't seem to align with the docs below:
> NODEID
> The unique ID for a node. A hex string that starts with an ! symbol.
>
> USERID
> A user ID string. This string is either a user ID if known or a nodeid to simply deliver the message to whoever the local user is of a particular device (i.e. person who might see the screen). FIXME, see what riot.im uses and perhaps use that convention? Or use the signal +phone number convention? Or the email address?
>
> CHANNELID
> FIXME, figure out how channelids work

Elsewhere on the page the term "DEVICEID" is used, is this synonymous with "NODEID"? With protobuf I get "msh/2/c" and with JSON I get "msh/2/json", so taking that all into consideration it feels like the format might be "msh/crypt/<serialization format>/CHANNELID/DEVICEID"?

I'm up for submitting a PR to make the page consistent, but not sure about the specific terms used.

garthvh:
The userid is the alpha numeric string you see, it maps to a node 1:1

PR #746, opened 2023-07-24
Update MQTT "Software Integrations" section

GUVWAF:
As noticed in #742 the MQTT “software integrations” section had some old/wrong information and it was missing quite some useful information, so I tried to add and correct it.

### DIFF
--- a/docs/software/mqtt/index.mdx
+++ b/docs/software/mqtt/index.mdx
@@ -25,22 +25,34 @@ IMPORTANT: When MQTT is turned on, you are potentially broadcasting your entire 
 
 ### MQTT Topics
 
-The device will uplink and downlink raw ([protobuf](https://developers.google.com/protocol-buffers)) packets to the `msh/` prefix:
+If no specific [root topic](/docs/settings/moduleconfig/mqtt#root-topic) is configured, the default root topic will be `msh/`.
+Each device that is connected to MQTT will publish its MQTT state ("`online`"/"`offline`") to:
 
-`msh/2/c/ShortFast/!12345678` where
+`msh/2/stat/USERID`, where `USERID` is the user ID of the gateway device (the one connected to MQTT).
 
-- `!12345678` is the address of the gateway device.
-- `ShortFast` is the channel name.
+For each channel where uplink and/or downlink is enabled two other topics might be used.
+
+#### Protobufs topic
+A gateway node will uplink and/or downlink raw ([protobuf](https://developers.google.com/protocol-buffers)) packets to the topic:
+
+`msh/2/c/CHANNELNAME/USERID`, where `CHANNELNAME` is the name of the channel.
+
+For example: `msh/2/c/LongFast/!abcd1234`
 
 The payload is a raw protobuf. Looking at the MQTT traffic with a program like `mosquitto_sub` will tell you it's working, but you won't get much useful information out of it. For example:
 
 ```text
 苓????"!
-	!937bed1cTanksTnk"D???05??=???aP`
-	ShortFast	!937bed1c
+!937bed1cTanksTnk"D???05??=???aP`
+ShortFast	!937bed1c
 ```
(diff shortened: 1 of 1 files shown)


docs/software/mqtt/index.mdx
```diff
@@ -90,34 +116,6 @@ Check out [MQTT Settings](/docs/settings/moduleconfig/mqtt) for full information
 `uplink_enabled` will tell the device to publish mesh packets to MQTT.
 `downlink_enabled` will tell the device to subscribe to MQTT, and forward any packets from there onto the mesh.
 
-### Topics
-
-The "mesh/crypt/CHANNELID/NODEID/PORTID" [topic](https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices) will be used for messages sent from/to a mesh.
-
-Gateway nodes will forward any MeshPacket from a local mesh channel with uplink_enabled. The packet (encapsulated in a ServiceEnvelope) will remain encrypted with the key for the specified channel.
```
tobymurray, line 97: Having just read through the legacy documentation, I think the references to "ServiceEnvelope" and a passing reference to "MeshPacket" were helpful. It lead me to a basic "where do I start" for interpreting the protobuf documentation.

docs/software/mqtt/index.mdx
```diff
@@ -25,22 +25,34 @@ IMPORTANT: When MQTT is turned on, you are potentially broadcasting your entire
...
+Each device that is connected to MQTT will publish its MQTT state ("`online`"/"`offline`") to:
 
-`msh/2/c/ShortFast/!12345678` where
+`msh/2/stat/USERID`, where `USERID` is the user ID of the gateway device (the one connected to MQTT).
 
-- `!12345678` is the address of the gateway device.
-- `ShortFast` is the channel name.
+For each channel where uplink and/or downlink is enabled two other topics might be used.
+
+#### Protobufs topic
```
tobymurray, line 35: It might be helpful to link early on to the Mestastic protobuf docs ( https://github.com/meshtastic/protobufs/blob/master/meshtastic) and mention that a bunch of this stuff is documented there?

Once directed to the protobuf docs, that feels like the "API documentation" in essence.

docs/software/mqtt/index.mdx
```diff
@@ -62,6 +74,20 @@ An example of a received `NODEINFO_APP` message:
 }
 ```
 
+The meaning of these fields is as follows:
+
+- "`id`" is the unique ID for this message. 
+- "`channel`" is the channel index this message was received on.
+- "`from`" is the unique node number of the node on the mesh that sent this message.
```
tobymurray, line 81: Worth mentioning the existing bug where in the JSON this intended-to-be-unsigned number is interpreted as a signed number (https://github.com/meshtastic/firmware/issues/2565)?

As this seems to be the go-to stable identifier for a specific node, it seems likely to be used and relied on. Seeing a different value via protobuf and JSON feels likely to yield confusion.

docs/software/mqtt/index.mdx
GUVWAF: Good points. I added it again in: https://github.com/meshtastic/meshtastic/pull/746/commits/ee40ecec4ddbaa3a125be2bf1b77fabd720c3c45

docs/software/mqtt/index.mdx
GUVWAF: Also added in https://github.com/meshtastic/meshtastic/pull/746/commits/ee40ecec4ddbaa3a125be2bf1b77fabd720c3c45

docs/software/mqtt/index.mdx
GUVWAF: Added in https://github.com/meshtastic/meshtastic/pull/746/commits/740ac50764f94fcfb5a40ece8c279366da524a3f

PR #777, opened 2023-08-11
MQTT docs first pass

pdxlocations:
This PR:
1. Removes a redundant paragraph
2. Emphasizes potentially broadcasting data to the public internet
3. Switch USERID references to NODEID
4. Change JSON forwarded packets  >> someone please check my work <<
5. Remove reference to uplink/downlink bug
6. A little of this and a little of that

[preview link](https://sandbox-git-mqtt-01-pdxlocations.vercel.app/docs/software/mqtt/)


GUVWAF:
Strictly speaking it actually is the `USERID` (namely [this field](https://github.com/meshtastic/protobufs/blob/master/meshtastic/mesh.proto#L467)), which is currently represented as the hexadecimal notation of the [Node number](https://github.com/meshtastic/protobufs/blob/master/meshtastic/mesh.proto#L976) with an exclamation mark in front.

Looking good further!

pdxlocations:
Thanks for having a look.  Do you think we should leave it as USERID then?  I was referencing the comments [here](https://github.com/meshtastic/firmware/blob/9470d4694bd5510174c8e0898af57cbd6f528c7f/src/mqtt/MQTT.h#L81C44-L81C44)

GUVWAF:
Yes, those comments appears to be wrong. Here's the actual code: https://github.com/meshtastic/firmware/blob/master/src/mqtt/MQTT.cpp#L472, where `owner` is from the `User` actually: https://github.com/meshtastic/firmware/blob/master/src/mesh/NodeDB.h#L31.

docs/software/mqtt/index.mdx
GUVWAF, line 1: Maybe we can add on line 91 that in firmware >=2.2.0 the JSON values are properly using unsigned.

pdxlocations: I added the note but I wonder if we could just say the values are unsigned since that is the current information.

pdxlocations: Actually I'm going to reword this a bit and change line 81 as well.  Hold off on review please.

pdxlocations: Reworded!  
[current preview](https://sandbox-git-mqtt-01-pdxlocations.vercel.app/docs/software/mqtt/#mqtt-topics)

PR #791, opened 2023-08-26
MQTT Public Server Connection Walkthrough

pdxlocations:
This Draft PR adds a framework for step-by-step instructions on how to connect to the default/public MQTT server. Comments/suggestions welcome!
[preview link](https://sandbox-git-mqtt-default-server-pdxlocations.vercel.app/docs/settings/moduleconfig/mqtt#connect-to-the-default-public-server)


pdxlocations:
> > > Are the tabs intentionally linked? When you click on a tab for the MQTT it'll change the one for the configuration above it also.
> > 
> > 
> > I guess I was thinking if you were interested in one category then it should stick but for this maybe not.
> 
> Hmm. I guess I get that. Although, personally, I don't like that you can't select an item separately. But that's just me

I was on the fence when setting it up but you've convinced me to make it independent. I'll switch it up in a minute.

PR #808, opened 2023-09-09
Smallerize images in MQTT walkthrough

pdxlocations:
The images are just a bit too big.  
Here's a [preview](https://sandbox-git-mqtt-smaller-images-pdxlocations.vercel.app/docs/settings/moduleconfig/mqtt#connect-to-the-default-public-server) of smaller ones

### DIFF
--- a/docs/configuration/module-config/mqtt.mdx
+++ b/docs/configuration/module-config/mqtt.mdx
@@ -130,8 +130,6 @@ All MQTT module config options are available for the Web UI.
   </TabItem>
 </Tabs>
 
-
-
 ## Connect to the Default Public Server
 
 <Tabs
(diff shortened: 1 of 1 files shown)


sachaw:
I think a better way of handling this is setting a max width globally for the rendered mdx. Keep it vanilla markdown if we can.

pdxlocations:
> I think a better way of handling this is setting a max width globally for the rendered mdx. Keep it vanilla markdown if we can.

I agree

It would be cool to be able to set the width by percentage or have a few options. 

I could also just size the images to fit properly in photoshop like I did here:
https://meshtastic.org/docs/software/android/usage

Honestly, for this PR I got lazy and pulled this sizing technique from:
https://meshtastic.org/docs/software/mqtt/nodered

pdxlocations:
> I think a better way of handling this is setting a max width globally for the rendered mdx. Keep it vanilla markdown if we can.

I resized the images and cleaned up the code.

PR #817, opened 2023-09-19
Add USB DFU to RAK4631 Bootloader update

linagee:
No J-Link cable is needed unless your bootloader is corrupted. Added documention to emphasize using USB DFU as the main bootloader update for RAK4631.

References:
- https://discord.com/channels/867578229534359593/871540140382761020/968759729667051550
- https://forum.rakwireless.com/t/bootloader-fails-to-upgrade-via-ble/4193/4


RicInNewMexico:
Might be worth adding a note to Step #5 that serial port naming may differ from the listed ttyACM0 depending on operating system. Other than that I think it looks good.

PR #820, opened 2023-09-20
Remote Hardware Module needs Python CLI

pdxlocations:
A note to make it clear that this module works through the Python CLI


pdxlocations:
Moving into modules

PR #831, opened 2023-09-29
More SerialModule updates

GUVWAF:
Found a left-over comment that it does not allow arbitrary protobuf commands yet, but it does now using the `PROTO` mode. Also updated the `TEXTMSG` mode, it will send all received text messages over the serial port as well: https://github.com/meshtastic/firmware/blob/master/src/modules/SerialModule.cpp#L286

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "clear": "docusaurus clear"
   },
   "dependencies": {
-    "@algolia/client-search": "^4.19.1",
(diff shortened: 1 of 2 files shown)


pdxlocations (approved):

pdxlocations:
@GUVWAF - Sorry I made some big changes to the directory structure and site-wide links before this got merged!

GUVWAF:
@pdxlocations No problem, will create a new PR.

PR #835, opened 2023-10-03
Add LONG_MODERATE to lora config

pdxlocations:


andrekir:
convention follows order in the protobufs, so your first commit was correct with [LONG_MODERATE](https://github.com/meshtastic/protobufs/blob/master/meshtastic/config.proto#L723) as the last value.

andrekir (approved):

PR #839, opened 2023-10-04
Not everyone's default channel is 20

pdxlocations:
Tips page said to use Channel 20 to join the "public" mesh.  This is not true for all regions.
Update QR codes to choose correct channel per region
[preview link](https://sandbox-git-tips-channel-20-pdxlocations.vercel.app/docs/configuration/tips)


pdxlocations:
> Can you make the list of channels a table instead? It'll be easier for users to quickly find their versus what is there now.

Ok

pdxlocations:
> > > Can you make the list of channels a table instead? It'll be easier for users to quickly find their versus what is there now.
> > 
> > 
> > Ok
> 
> If you don't want to, it's fine. I just find it's much easier to find the specific information I need for my region in the list of 15 different regions, when it's in table format. That's just me though. Happy to do whatever is most aesthetically pleasing.

 I would prefer a table too, but our tables are just so ugly, I'd like to leave it as a list for now.  
 I tried it as a vertical table and a horizontal table but they just didn't work for me.  I'd make a pretty picture but then nobody would be able to edit it if there's a mistake or new regions are added.  Thanks for your understanding.

PR #858, opened 2023-10-27
[ADD] description to enter service mode

davudevren:
I struggled a bit with flashing meshtastic firmware to the t-beam s3 supreme (shipped with SoftRF).
After i enter the service mode i was able to flash the firmware. I think some information in the wiki about that would be helpful.


pdxlocations:
> I struggled a bit with flashing meshtastic firmware to the t-beam s3 supreme (shipped with SoftRF). After i enter the service mode i was able to flash the firmware. I think some information in the wiki about that would be helpful.

@davudevren could you sign the CLA when you get a chance?  This seems to be a useful contribution, though I don't have a supreme to verify this.

caveman99:
@davudevren could you please sign the CLA? We can't merge this otherwise.

PR #859, opened 2023-10-30
Update index.mdx

wschnell:
I asked the author, shall I (try to) put a link to the Telegram gateway into the official Meshtastic documentation? he wrote:
That would be awesome, yes. Keep in mind that even it works, it's far from being perfect :) With kind regards, Bohdan Turkynewych


caveman99:
@wschnell can you please have a look at the comments?

PR #863, opened 2023-11-03
Add instructions for measuring power in new Solar Powered category

KeithMon:
This document is the first in a series of solar-specific documents.


pdxlocations:
@KeithMon wow, nice work!  I read through it's very well written.  
@rcarteraz I think having this as a sub-page will ultimately make sense when there are more pages.  It seems like Keith is planning to add a "Panel Size" page shortly. Right?

pdxlocations (approved):

pdxlocations:
We may eventually want an index.mdx page to land on with some basic recommendations: nRF, soshine, green power module, etc.

KeithMon:
@pdxlocations @rcarteraz thanks guys. Yes, this is the first in a series of documents. Next will be about calculating solar panel and battery bank size. It will eventually make sense to have an index page.

pdxlocations:
> @pdxlocations @rcarteraz thanks guys. Yes, this is the first in a series of documents. Next will be about calculating solar panel and battery bank size. It will eventually make sense to have an index page.

Do you want this one merged now, or wait until "Save your test result for the next step, calculating solar panel size." is ready?

KeithMon:
@pdxlocations If you think it's best to wait we can do that. I don't have a lot of time to dedicate to this so I don't know when I'll be able to get the solar panel page uploaded. I would prefer to take this one live now and I'll add more as I'm able. The solar panel doc exists publicly right now (pinned in Discord) so it shouldn't take much to prepare it for the site.

PR #875, opened 2023-11-10
Corrected String lengths for SSID and NTP server name

tobi-sys:
According to the source in in /firmware/src/mesh/generated/meshtastic/config.pb.h  the declarations are
char wifi_ssid[33];
char ntp_server[33];
so these variables can hold 32 characters and the terminator \0


tobi-sys:
relates to the pull request https://github.com/meshtastic/protobufs/pull/402

caveman99 (approved):

PR #906, opened 2024-01-04
Clarify step 4 a bit in the private primary/default secondary tip

guillochon:
I think users would reasonably assume that the LoRa channel is by default set to their region's channel and are unaware of the default hashing algo used to generate the channel number. Rewriting this way makes it less likely for users to make this mistake.

### DIFF
--- a/docs/configuration/tips.mdx
+++ b/docs/configuration/tips.mdx
@@ -29,7 +29,7 @@ If you'd like to connect with other Meshtastic users but only share your locatio
 1. Ensure you have not changed the LoRa [Modem Preset](/docs/configuration/radio/lora#modem-preset) from the default `unset` / `LONG_FAST`.
 2. On your PRIMARY channel, set anything you'd like for the channel's name and choose a random PSK.
 3. Enable a SECONDARY channel named "LongFast" with PSK "AQ==".
-4. Since the radio's frequency is automatically changed based on your PRIMARY channel's name, you will have to manually set it back to your region's default (in LoRa settings) in order to interface with users on the default channel:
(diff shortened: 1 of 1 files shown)


docs/configuration/tips.mdx
```diff
@@ -65,4 +65,4 @@ This configures the frequency the radio is set to.  Check out the [frequency cal
 - If you are part of a large mesh and don't know what a setting does, don't change it (unless you're super curious).
 - Leave your [MAX HOPS](/docs/configuration/radio/lora#max-hops) set to 3 unless you're sure you need more (or less) to reach your destination node.
 - TEST your settings and hardware before you install in hard-to-reach locations.
-- Connecting a node to the [public MQTT server](http://localhost:3000/docs/configuration/module/mqtt#connect-to-the-default-public-server) may publish the locations of all nodes in your mesh to the internet.  This will also add every globally connected node to your node database and potentially flood your mesh with all types of packets.
\ No newline at end of file
+- Connecting a node to the [public MQTT server](http://localhost:3000/docs/configuration/module/mqtt#connect-to-the-default-public-server) may publish the locations of all nodes in your mesh to the internet.  This will also add every globally connected node to your node database and potentially flood your mesh with all types of packets.
```
pdxlocations, line 68: looks like I made a mistake with this link too.  I believe it should be:
./module/mqtt#connect-to-the-default-public-server

guillochon: I'm not sure what you'd like the link to be changed to. Also I'm not sure what's going on with the diff here...I don't see any changes to this line

pdxlocations (approved):

docs/configuration/tips.mdx
pdxlocations: More of a mental note.  I'll fix it later.  Not sure why it came up in the diff, but I'm glad it did!

PR #912, opened 2024-01-05
Spell out explicitly what channel config options need to match

guillochon:
Currently the settings that need to match is only implicitly specified in the docs on the tips page. While you can infer that channel names need to match for communication to work, it seems like it'd be easier for new users to understand what's going on by adding that information here in the initial config pages.


guillochon:
Oh I had missed that, thanks @rcarteraz, appreciate it.

PR #915, opened 2024-01-10
Add T3S3 to list of SX128X's in audio module config

pdxlocations:


sachaw:
Yeah, that's normal, it's just listing/formatting checks.

PR #929, opened 2024-01-20
Add local groups page

crypto-smoke:


crypto-smoke:
I figured new users that are getting started would like to connect with local groups. I also couldn't find a place that felt better. Open to whatever though.

crypto-smoke:
I like the community idea. It would be a good category for other community stuff, like an "awesome Meshtastic" page for cool things people have done or created. Maybe the distance records could move there too?

pdxlocations:
I wouldn't be opposed to a whole new 1st level category.  "Community"?  I could see a map in there eventually, maybe hardware/software builds.  We could move the 'Community Apps' out of Software.  Actually some of those are now 'official' Meshtastic software...

One thing though I would definitely check with the groups that you're linking to before adding them.  I don't see why they'd object but they may want to "clean up" first.

PR #943, opened 2024-01-25
Move MQTT into Software Integrations

pdxlocations:
[preview link](https://sandbox-git-mqtt-integrations-pdxlocations.vercel.app/docs/software/integrations)


vercel.json
```diff
@@ -71,6 +71,14 @@
     {
       "source": "/docs/legal/licensing/",
       "destination": "/docs/legal/licensing-and-trademark"
+    },
+    {
+      "source": "/docs/software/mqtt",
```
tekstrand, line 76: We should probably consider turning on trailing slash so we don't have to deal with stuff like this and it's better for SEO

https://vercel.com/docs/projects/project-configuration#trailingslash

tekstrand (approved):
Changes look good to me. Trailing slash recommendation should be it's own PR.

pdxlocations:
> Looks like some conflicts will need to be resolved after merging @GUVWAF 's PR. 

Resolved, and turns out it was one of my PRs conflicting.

PR #945, opened 2024-01-25
Standardize hardware documentation

tekstrand:
## What
1. Update headings to follow semantic guidelines
2. Shorten some tab names (t-beam)
3. Add missing headings under tabs for semantics
4. Remove extra duplicate words from headings when parent heading already contains context
5. Standardize purchase link formatting for all RAK devices

## Why
1. Semantic html is easier to read, better for accessibility, and better for SEO
2. There were enough tab names and the names were long enough that they were starting to wrap

## Screenshots
Screenshot 2024-01-26 at 1 06 16 PM
Screenshot 2024-01-26 at 1 06 10 PM
Screenshot 2024-01-26 at 1 06 06 PM
Screenshot 2024-01-26 at 1 06 02 PM
Screenshot 2024-01-26 at 1 05 59 PM

PR #947, opened 2024-01-26
Update peripherals.mdx for RAK19007+RAK12500 combination

ianmcorvidae:
a RAK12500 plugged into slot A (they only fit, physically, on A or C, looks like) appears to be working on my newly-setup RAK19007/RAK4631 device, so I added the 19007 to the section on using a RAK12500


docs/hardware/devices/rak/peripherals.mdx
```diff
@@ -29,7 +29,7 @@ Further information on the RAK1910 can be found on the [RAK Documentation Center
 
 US Distributor - Purchase link: [Rokland](https://store.rokland.com/products/rak-wireless-rak1910-wisblock-gnss-location-module)
 
-To add a GPS to the RAK19003 base board, you need the [RAK12500 GPS sensor](https://store.rakwireless.com/products/wisblock-gnss-location-module-rak12500). It is supported via I<sup>2</sup>C on slot B for firmware versions 1.49 and above.
+To add a GPS to the RAK19003 or RAK19007 base boards, you need the [RAK12500 GPS sensor](https://store.rakwireless.com/products/wisblock-gnss-location-module-rak12500). It is supported via I<sup>2</sup>C on slot B for firmware versions 1.49 and above.
```
pdxlocations, line 32: Hmm, the RAK1910 is also supported on the 19003 and 19007.  I can't remember off the top of my head what slots these use.

ianmcorvidae: Looking at the RAK docs, it looks like A as well, maybe. I realize my docs update is wrong anyway, since it says to use slot B. And I'm not sure that's right either, the RAK docs seem to suggest the 19003 has only slots C and D (and it looks like C is the one that's the right size for either GPS module), no A or B at all. So something's amiss, anyway.

ianmcorvidae: I updated it to say that you _should_ get the RAK12500 (since the RAK1910 on their site suggests the RAK12500 as an updated replacement anyway, that seems correct, and corrected my notes about which slot (A) to use on the RAK19007. I don't really know the correct answers for the other boards, so I haven't touched those.

tekstrand:
This one is going to have some conflicts with #945 I believe

pdxlocations:
I believe it would be more appropriate to list all supported hardware and make a note that the 12500 is newer and perhaps list it's advantages.  The 1910 is still widely available and a bit cheaper so we don't want people thinking it's not supported.  Sorry, this page was a bit of a mess to start.

ianmcorvidae:
That's totally fair, I've just been cautious about adding things I'm unsure about. Would it be better for me to build something on top of #945? I guess I'm at least confident enough to state that RAK12500 and RAK1910 are both compatible with RAK19007 & 19003 so I could rewrite the section to address that, but I'm unsure if 945 is likely to/needs to go in first or anything like that.

pdxlocations:
> but I'm unsure if 945 is likely to/needs to go in first or anything like that.

#945 has been merged

there was some conversation on GPS chips and slots today:
https://discord.com/channels/867578229534359593/919642584480112750/1201209718241230888

If you search the discord for 1910 and 12500 you'll find all the answers.  A lot of things have changed over the past year or so.

ianmcorvidae:
Alright, I've updated this and reorganized it slightly. The page headings are now for the GPS modules rather than for combinations of GPS/base board, which matches the other peripherals sections, and under each the supported boards & slots are listed. As far as I can tell, the 5005 board only supports the 1910. I removed the details about I2C/UART as they don't seem relevant to someone buying hardware (and everything uses UART now, according to discord).

pdxlocations (approved):
Looks great to me

PR #951, opened 2024-01-29
Update MQTT JSON downlink according to new requirements

GUVWAF:
In https://github.com/meshtastic/firmware/pull/3141, the ability to send messages on the mesh by crafting a JSON message and send it on an MQTT server got some stricter requirements (to avoid things like [this](https://meshtastic.discourse.group/t/strange-repeating-messages/9548)). I therefore updated the documentation accordingly and left a comment about older firmware. 

I hope I didn't miss a spot. The example images on the NodeRed page are now outdated, but I think the accompanying text makes it clear.


garthvh (approved):
Looks really good

docs/software/mqtt/index.mdx
```diff
@@ -96,19 +96,25 @@ The `from` field can thus be used as a stable identifier for a specific node.  N
...
+   "type": "type",
    "payload": {
        "key":"value"
        ...
     }
 }
 ```
 
-`sender` and `payload` fields are required for a valid envelope. If a valid MQTT message is found, the message is sent over the radio as a message of type `TEXT_MESSAGE_APP` with the serialized `payload` value in the message payload.
+`from` and `payload` fields are required for a valid envelope (note that in firmware <2.2.20 a field `sender` was required, but this is no longer the case). The `from` field should be equal to the node number of the node that will transmit the message. Optionally, you can specify a destination by setting the `to` field to the node number of the destination. If the `to` field is not set, the message will be broadcast to all nodes on the mesh.
```
pdxlocations, line 114: This line is causing the build to fail, you might try '&lt; 2.2.20'

GUVWAF: Ah, sorry, missed that. Changed it and checked that it builds now.

tekstrand:
If this merges first will require update of #943

PR #954, opened 2024-01-30
Update device.mdx

Nestpebble:
update LOST_AND_FOUND text to make it clear that the messages are sent on the default channel for the device.


Nestpebble:
This is wrong - should be withdrawn.

PR #955, opened 2024-01-30
Update device.mdx

Nestpebble:
clarified the channel that the LOST_AND_FOUND role uses for messages.


docs/configuration/radio/device.mdx
```diff
@@ -28,7 +28,7 @@ Acceptable values:
...
-| `LOST_AND_FOUND` |                                                                                                                                                                                                  Lost and Found - Used to automatically send a text message to the mesh with the current position of the device on a frequent interval: "I'm lost! Position: lat / long"                                                                                                                                                                                                   |
+| `LOST_AND_FOUND` |                                                                                                                                                                                                  Lost and Found - Used to automatically send a text message on the primary channel for the device with the current position of the device on a frequent interval: "I'm lost! Position: lat / long"                                                                                                                                                                                                   |
```
tekstrand, line 31: Just being pedantic but I think `to` was better here.

tekstrand (approved):
Changes lgtm 🍺

pdxlocations (dismissed):

docs/configuration/radio/device.mdx
pdxlocations, line 31: I was about to merge this but realized you use 'device' twice.  Would you consider:

"Used to automatically send a text message with current position at frequent intervals to the primary channel for the device: "I'm lost! Position: lat / long""

Nestpebble: > Just being pedantic but I think `to` was better here.

I'm not fussed either way - I just wanted it clear that it goes on the primary channel!

Nestpebble: Yes, I think I've changed the pull request to this language. Thanks for making it clearer.

pdxlocations (approved):

PR #957, opened 2024-02-01
Add MQTT to integrations index

pdxlocations:


pdxlocations:
I had this in there but it got lost somewhere along the line.

PR #958, opened 2024-02-01
Trailing Slash = True

pdxlocations:
Attempts to clean up the forwards by forcing trailing slash

Issue is it isn't working with forwards using :path*
for example
 /docs/settings/moduleconfig/audio won't forward to /docs/configuration/module/audio/

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -12,7 +12,7 @@
     },
     {
       "source": "/docs/settings/config/:path*",
-      "destination": "/docs/configuration/radio/:path*/"
+      "destination": "/docs/configuration/radio/:path*"
     },
     {
       "source": "/docs/settings/moduleconfig/",
--- a/vercel.json
+++ b/vercel.json
@@ -1,8 +1,6 @@
 {
+  "trailingSlash": true,
   "redirects": [
-    {
-      "trailingSlash": true
-    },
     {
       "source": "/docs/settings",
       "destination": "/docs/configuration"
--- a/vercel.json
+++ b/vercel.json
@@ -1,17 +1,10 @@
 {
   "trailingSlash": true,
   "redirects": [
-    {
-      "source": "/docs/settings",
-      "destination": "/docs/configuration"
-    },
+    
     {
       "source": "/docs/settings/",
       "destination": "/docs/configuration/"
-    },
-    {
-      "source": "/docs/settings/config",
-      "destination": "/docs/configuration/radio"
     },
     {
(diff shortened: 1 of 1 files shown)


vercel.json
```diff
@@ -20,24 +14,20 @@
       "source": "/docs/settings/config/:path*",
       "destination": "/docs/configuration/radio/:path*"
     },
-    {
-      "source": "/docs/settings/moduleconfig",
-      "destination": "/docs/configuration/module"
-    },
     {
       "source": "/docs/settings/moduleconfig/",
```
tekstrand, line 18: I think `/audio` isn't forwarding because you don't have `path` here. If you need subpaths here to forward you'll have to add that.

vercel.json
```diff
@@ -20,24 +14,20 @@
...
       "destination": "/docs/configuration/module/"
     },
     {
       "source": "/docs/settings/moduleconfig/:path*",
-      "destination": "/docs/configuration/module/:path*"
+      "destination": "/docs/configuration/module/:path*/"
```
tekstrand, line 23: this is just `module` here and not `moduleconfig` another possible issue?

pdxlocations: the base paths for radio and module settings were/are

      "source": "/docs/settings/config/",
      "destination": "/docs/configuration/radio/"
      
      and
      
      "source": "/docs/settings/moduleconfig/",
      "destination": "/docs/configuration/module/"

vercel.json
pdxlocations: doesn't work with trailingslash=true
https://sandbox-71lgde8mm-pdxlocations.vercel.app/docs/settings/moduleconfig/audio

works
https://meshtastic.org/docs/settings/moduleconfig/audio

tekstrand: I'm not sure how this is happening because I don't see it being removed in this PR, but [L#32](https://github.com/meshtastic/meshtastic/blob/master/vercel.json#L32) of the config currently has a redirect for `moduleconfig/:path` while yours does not. I think that's why it's not working. This redirect still needs to be somewhere in your vercel.json.
```
    {
      "source": "/docs/settings/moduleconfig/:path*",
      "destination": "/docs/configuration/module/:path*"
    },
````

pdxlocations: I tried a bunch of variations so it may have not been changed back.  I removed a slash that I added.

tekstrand:
I opened #959 where I tried making the changes to see what would happen

PR #974, opened 2024-02-07
enforce trailing slash in docusaurus

tekstrand:
## What
We turned on trailing slashes in #958 but missed the docusaurus config which adds the trailing slash to each link on the page

## Why
We need consistency to avoid duplicate pages in Search Engines

## Test
Build/run locally and check navigation links, they should all end in a trailing slash


pdxlocations (approved):
I haven't tested this but if it's working for you, I trust you.

PR #975, opened 2024-02-07
301 old faq page to new

tekstrand:
## What
Adds a 301 for the old FAQ page to the new location

## Why
404s are bad for users :)

## Test
Visit /docs/getting-started/faq and you should be redirected to /docs/faq/


pdxlocations (approved):
finally, thank you.

PR #976, opened 2024-02-07
Add triple press button action for heltec

5kamisama5:
Just adding the triple click function.


tekstrand (approved):
Changes look good to me.

PR #980, opened 2024-02-07
Update core-modules.mdx

jopdyke:
added US distributor for RAK11310


tekstrand (dismissed):
We have US first on all the other pages before international, would you mind adding US above international here please?

PR #1005, opened 2024-02-13
Update power.mdx

nagumii:
Removed the specific reference to lipo batteries, as instructions also apply to lithium ion batteries such as 18650s.


pdxlocations:
> Oops. @pdxlocations sorry you must have hit submit right before me.

https://www.youtube.com/watch?v=0rpdQvuAnqM

PR #1010, opened 2024-02-15
faq click sets hash in url for scrolling

tekstrand:


tekstrand:
@pdxlocations @rcarteraz here is a working "scroll to last clicked item". Only problem I am seeing is that every time you open an FAQ item it scroll it to the top of the page and feels very weird. What do y'all think?

pdxlocations:
> @pdxlocations @rcarteraz here is a working "scroll to last clicked item". Only problem I am seeing is that every time you open an FAQ item it scroll it to the top of the page and feels very weird. What do y'all think?

oh, I don NOT like that.

tekstrand:
I agree. What do y'all think the ideal behavior is?

pdxlocations:
it's weird that these links send  you halfway down the page
https://meshtastic.org/docs/faq/?#modules

tekstrand:
Maybe we can add a specific link button like headings have? You can click it, it updates the url, and then you can copy that?

PR #1013, opened 2024-02-15
Update network.mdx

nagumii:
Adds mention that enabling wifi disables bluetooth.


nagumii:
My mistake, it's already mentioned.

PR #1034, opened 2024-02-20

moresecode:
Updated: Store & Forward > Client Config > Apple INFO.

### DIFF
--- a/docs/configuration/module/store-and-forward-module.mdx
+++ b/docs/configuration/module/store-and-forward-module.mdx
@@ -118,7 +118,10 @@ Store and Forward Config options are available for Android.
 <TabItem value="apple">
 
 :::info
-Store and Forward configuration is not currently available via the Apple clients.
+Store and Forward Config options are available for iOS client.
+
+1. Open the Meshtastic App
+2. Settings > Module Configuration > Store & Forward
 :::
 
 </TabItem>


docs/configuration/module/store-and-forward-module.mdx
```diff
@@ -118,7 +118,10 @@ Store and Forward Config options are available for Android.
 <TabItem value="apple">
 
 :::info
-Store and Forward configuration is not currently available via the Apple clients.
+Store and Forward Config options are available for iOS client.
```
tekstrand, line 121: We'll need to update this to be like the other Modules. They say something like

>All Ambient Lighting Module config options are available on iOS, iPadOS and macOS app versions 2.2.3 and higher at Settings > Modules > Ambient Lighting

@garthvh do you know what version S&F was added to for the Apple apps?

garthvh: 2.2.3 and greater is correct, it is somewhat hard to use older App Store versions

tekstrand: @moresecode can you please update your text to match the other sections? Thanks!

moresecode: But I'm using version 2.2.23 on my phone.

PR #1037, opened 2024-02-22
Update home-assistant.mdx to add device_class and use this.state

ianmcorvidae:
Just a couple small improvements. By adding these device classes, some history graphs will be made a bit nicer (splitting out battery percentage from channel utilization graphs) and HA can better know how to use the values. Using `this.state` is slightly better form for retaining the same sensor value since it won't look up the current value, just reuse the one HA has already looked up.

### DIFF
--- a/docs/software/integrations/mqtt/home-assistant.mdx
+++ b/docs/software/integrations/mqtt/home-assistant.mdx
@@ -46,7 +46,7 @@ sensor:
             value_json.payload.temperature is not defined %}
         {{ (value_json.payload.voltage | float) | round(2) }}
       {% else %}
-        {{ states('sensor.node_1_battery_voltage') }}
+        {{ this.state }}
       {% endif %}
     unit_of_measurement: "Volts"
     # Telemetry packets come in two flavors: The default node telemetry, and the I2C sensor data.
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
I also added some notes about units. An upside of using these device classes on top of the metadata in HA allowing nicer data display is that units are automatically converted to your preferred HA unit for that device class. Unfortunately, that behaves poorly with the `else` conditions that make the sensor retain the same value on irrelevant messages -- if you set a Celsius unit for temperature but HA is configured to use Fahrenheit, then when actual temperature readings come in the value is set correctly, but when non-temperature messages get processed, the value from `this.state` (or from `states('...')` as it was before) will be the value in Fahrenheit but will be _interpreted_ as a Celsius temperature. So if 12C sets the value as 53.6F, then on the next non-temperature message the value would change to 128.48F (i.e., 53.6C). Then 263.264F, 505.8752F, etc., until the next actual temperature reading comes in. The other fix to this would be to look at the unit in the sensor attributes on the `this` object and convert the value as appropriate if needed, but it's probably easier to just match the units that HA would convert it into. Also applies to pressure, so the note is there too.

GUVWAF (approved):
I'm not a Home Assistant user myself (I use OpenHAB), but I trust the author on this as the explanation sounds good.

Issue #226, opened 2022-02-03
PDF Handbook Generator


nuxnik:
# Synopisis

We should offer an PDF version of the documentation which users can download and view offline. Here are some requirements:

1. The PDF should be based on the status of the source files of the current documentation 
2. A new release of the documentation should coincide with a newly generated PDF
3. Generating a document should be based on some sort of Github action trigger
4. The document artifacts should in turn be made available for download to the documentation website: https://meshtastic.org
5. The document should have a cover sheet complete with current version number and if possible a table of contents

Documentation can be successfully generated with the following console commands. See [Docusaurus Prince PDF](https://github.com/signcl/docusaurus-prince-pdf):
```bash
npx docusaurus-prince-pdf -u https://meshtastic.org/docs/software --output meshtastic-docs.pdf
```
I was thinking the easiest way of doing this would be to put all the build tools into a docker image. This way, all the project dependencies will be encapsulated in the container and can be executed in any environment. In the container, the project would be built, spun up, then an internal process in the container could build and generate the document with Docusaurus Prince PDF. 

I'm a noob with GitHub Actions, but have a lot of experience in configuring GitLab CI Pipelines. Is is possible to trigger a docker container within GitHub or does this have to happen on an external server?

jfirwin:
Looks pretty awesome. I'd love a table of contents added as we talked about on discord.

nuxnik:
@jfirwin I'm looking into updating the library to create a TOC. PR is in the works.

jfirwin:
Awesome! I'm excited to see what you come up with. Is the TOC built using `sidebar.js`? That might be a good way to break up sections dynamically.

nuxnik:
Good point. That seems like the best place to get a complete outline of the docs. As it is, the script crawls whatever starting point you set then follows all "next" links. Now that you mention it, I might end up cannibalizing the whole script and writing a new generator that reads from the sidebar.js.

jfirwin:
I saw some interesting comments on one of the few docusaurus issues related to pdf generation. They mentioned the sidebar.js and that's what got me thinking about it.

https://github.com/facebook/docusaurus/issues/969

nuxnik:
Ah yes, that is true. Ideally, all tabs should be rendered into the documentation and added to the TOC. The selector in the generator script searches for CSS components. I suppose I could add another selector to the script which will expand the other tabs and parse the data. Haha, it's getting more and more complicated. Good call :)

[...]

PR #1041, opened 2024-02-24
FAQ accordion open sets nearest heading hash in url

tekstrand:
## What
- Updates the FaqAccordion component so that when an accordion is opened, the hash of the nearest heading is added to the url of the page 
    - This **no longer** automatically opens accordion items for the user
- Remove `slug` from Faq.mdx because it's no longer needed 🥳 

## Why
- Better for sharing, auto scrolls user to the right section
- Handling auto opening was extremely complex and caused an issue where we had to wait for client side rendering to handle FAQs. This causes other issues like generating PDFs of the docs is not possible.

### DIFF
--- a/docs/about/faq.mdx
+++ b/docs/about/faq.mdx
@@ -190,38 +190,38 @@ If you use your ham radio license with Meshtastic, consider both the privileges 
 
 ## Overview
 
-<FaqAccordion rows={Faq.general} slug="general" />
+<FaqAccordion rows={Faq.general} />
 
 ## Android Client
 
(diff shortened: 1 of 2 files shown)


tekstrand:
@pdxlocations and @rcarteraz finally got around to a solution that fixes the concerns y'all raised with the auto scrolling/link sharing on the FAQs. Let me know what y'all think of this solution :)

This, in combination with #1021 sets us up nicely to be able to finally solve https://github.com/meshtastic/meshtastic/issues/226

PR #1042, opened 2024-02-24
Work In Progress: Automatic Documentation PDF Generation

tekstrand:
## What
Solves #226 
1. Adds `docusaurus-prince-pdf` as dev dependency
2. Some new css that makes tabs hidden but their content visible when printing pages
3. A new GitHub action that
    - Installs [Prince xml](https://www.princexml.com/)
    - Generates a PDF
    - Fetches access token from Meshtastic org github app
    - Pushes the generated pdf to the `docs` branch of `meshtastic/artifacts` as `Meshtastic-Documentation-${DATE_TIME}_${{ github.sha }}.pdf`

## Why
- We have users who would like to maintain offline documentation
- We have [users with brain injuries](https://www.reddit.com/r/meshtastic/comments/1au18gr/offline_manual/) who have a hard time reading our docs on screen
- All tabs now have headings after #1021 so the actual tab can be hidden and the content kept visible

### DIFF
--- a/.github/workflows/pdf.yml
+++ b/.github/workflows/pdf.yml
@@ -1,12 +1,25 @@
 name: Generate Documentation PDF
 on:
-  schedule:
-    - cron: '0 0 * * *'
+  # # Triggers the workflow on push but only for the master branch
+  push:
+    branches: [master]
+
+  # Note: This is different from "pull_request". Need to specify ref when doing checkouts.
+  pull_request_target:
+    branches: [master]
+    paths-ignore:
+      - "**.md"
 jobs:
   build:
     if: github.repository == 'meshtastic/meshtastic' && github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


caveman99 (approved):

tekstrand:
@sachaw updated to match firmware action. Not sure how to test it in this configuration

sachaw:
Just trying to see if I can make the `ARTIFACTS_TOKEN` an org level token

tekstrand:
@sachaw @thebentern seems like this workflow skipped the artifacts step because the criteria was not met
https://github.com/meshtastic/meshtastic/blob/master/.github/workflows/pdf.yml#L33
`if: ${{ github.event_name == 'pull_request_target' || github.event_name == 'pull_request' }}`

I copied this directly from the firmware builds as suggested. What is the workflow over there that successfully triggers it?

tekstrand:
This should fix it https://github.com/meshtastic/meshtastic/pull/1063

PR #1049, opened 2024-02-26
Add Donate Button to Nav Bar

rcarteraz:
What this PR does:
- Adds a donate button to the Nav Bar. Added to the left side as the right side is collapsed when viewing on smaller screens. Still partially blocked on smaller screens in vertical mode but most visible and selectable. 
- Removes the title from the Nav Bar. This was redundant imo, and mostly cutoff on mobile devices anyways. This makes it looks cleaner.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -23,7 +23,6 @@ const config = {
       },
     },
     navbar: {
-      title: "Meshtastic",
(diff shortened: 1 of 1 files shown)


tekstrand (approved):
Changes look good to me!

PR #1053, opened 2024-02-26
Create meshtastic-coverage-mapping.mdx

Adrelien:
Adding a nice free tool to make mapping coverage 
from my blog 
https://adrelien.com/blog/how-to-do-coverage-mapping-meshtastic/


tekstrand:
Agree it doesn't make sense under software. Guides seems good. I was thinking something like that for when this was added the other week: https://meshtastic.org/docs/hardware/devices/heltec/peripherals/

Having a section like this could make people more likely to contribute. I think the counter argument is that guides can tend to get out of date faster, and having them on the official website would bring assumptions from readers that it is "endorsed" by Meshtastic, or something like that. Mainly thinking ahead to future contributions as I don't think anything we have currently falls under that.

pdxlocations:
This is a larger conversation.  I mean the Home Assistant, NodeRED, etc are also community projects.  It would be easier in the long run to have all of these associated projects in other's personal blogs, hack-a-day, etc.

I like having all these project live under the .org but it will be a maintenance nightmare when there are changes in the firmware.  This just happened for the HA page when there was a change.

We might want to discuss on Discord so others can chime in before making a decision.

Or just go with community/guides and add a disclaimer that the projects aren't officially endorsed by Meshtastic.

tekstrand:
I like this, it was going to be my recommended approach if we chose to go forward.

It could be interesting, in the long run, to install a blog platform alongside the docs on the website, and let folks guest write. Having more content on .org is fantastic for SEO, plus authors would want to help drive traffic.

tekstrand:
Also, I know it is a contentious topic, but

>We might want to discuss on Discord so others can chime in before making a decision.

RFCs are a perfect place for this

Adrelien:
I like the idea of separate blog and people guest blog. Something like ghost what I am running it nice. It’s also nice for the writer to get some credit. Like for me It takes time and tools to research and do something then when I publish it I would like it if I get a backlink. Don’t get me wrong I am fine giving it for free still but a backlink would be nice motivation to dive more into the topic. At the end I am an engineer with writing hobby would like to see myself moving in the right direction.

tekstrand:
>docusaurus offers it 

Cool! I don't personally feel like we're that close to moving away from Docusaurus as I haven't heard anyone talk about it seriously since I've been around(admittedly, only 3 months).

tekstrand:
Man that actually looks awesome. Nice job @rcarteraz!

I _think_ I lean towards one blog, until we have a problem of too many blog posts in one area, which I don't think we will have yet. We can use the `tag` system they offer to have a community post/meshtastic post URL, I think. 

What do y'all think?

docs/community/software/meshtastic-coverage-mapping.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+### Define Your Site:
+
+*   Log in, go to the main menu, and select "New Site."
+*   Place a pin or enter coordinates for your site.
+*   Click submit, name your spot, add elevation, and hit "Add To My Sites."
+
+### Building Coverage Maps:
+
+![](https://adrelien.com/blog/content/images/2024/02/meshtastic-map-settings-2.png)
```
wheresalice, line 38: Referencing an external image seems like a bad idea

docs/community/software/meshtastic-coverage-mapping.mdx
```diff
@@ -0,0 +1,95 @@
...
+Discover the power of Meshtastic with our step-by-step guide on creating coverage maps for your Lora nodes. These maps are key in optimizing communication range, visually showcasing signal strength and coverage areas within a LoRa mesh network.
+
+In this guide we will use a really cool tool to create coverage mapping for our Meshtastic Lora node. Coverage maps play a crucial role in optimizing communication range for Meshtastic devices. These maps visually represent the signal strength and coverage areas of a LoRa mesh network, offering valuable insights into the performance of your devices
+
+Getting Started:
+----------------
+
+### Create Your Account:
+
+*   Visit [https://www.ve2dbe.com/rmonline\_s.asp](https://www.ve2dbe.com/rmonline_s.asp?ref=adrelien.com).
```
tekstrand, line 27: We shouldn't be adding `ref` attributes to other websites.

tekstrand:
I agree with this.

PR #1062, opened 2024-02-29
Add documentation download link

tekstrand:


tekstrand:
Thanks. Once we have the PDF actually working we can test this one out

pdxlocations:
The word "scraping" is a little abrasive.  Could I suggest something along the lines of 
"Current Meshtastic.org documentation compiled for offline use"

This is amazing by the way.  Nice work!

tekstrand:
♻️ updated the copy.

@rcarteraz I like the idea of adding it to the API so we can link directly to the latest, but it's definitely a heavier lift. Are you okay if we leave it for now and when I get more time I can follow up with that?

PR #1066, opened 2024-03-01
bump prince pdf version to 15.3 latest

tekstrand:
## What
Bumps prince (pdf generator library) to version 15.3

## Why
webp support was added in 15 https://www.princexml.com/releases/15/


tekstrand:
Yeah i'm an idiot https://github.com/meshtastic/meshtastic/pull/1068

tekstrand:
and back to that same 403 error. May have to troubleshoot later w/ someone who has admin permissions

PR #1071, opened 2024-03-01
Channel Number to Frequency Slot

### LINKED ISSUE #1017: Rename Lora Channel to Frequency Slot
In order to have a single set of channels, we are going to re-name the lora channel to frequency slot

https://github.com/meshtastic/protobufs/blob/5f28be497a5518334c86378335e8ffcd177ed661/meshtastic/config.proto#L878

We will update the name of the the protobuf value in 3.0

rcarteraz:
### What

Changes the channel number to frequency slot. Closes #1017 

### Why

Channel number under the LoRa settings often gets confused with the channels used for communication. The decision was made to update this to "frequency slot" to avoid the confusion. All reference to "channel number" for the lora setting will be changed to "frequency slot". In the next breaking change the Protobuf will also be changed to match.


tekstrand (approved):
Changes lgtm. New description is much easier to read 👍

thebentern (approved):

PR #1083, opened 2024-03-05
improve title/meta description for sharing and SEO

rcarteraz:
### What

Updates the title for the MQTT page and  adds a meta description.

### Why

Original lacked hardly any info other than "MQTT"

Screenshot 2024-03-05 at 12 11 04 PM

### DIFF
--- a/docs/software/integrations/mqtt/index.mdx
+++ b/docs/software/integrations/mqtt/index.mdx
@@ -1,6 +1,6 @@
 ---
 id: mqtt
-title: MQTT | MQTT Integrations Overview
+title: MQTT | Integrations Overview
 sidebar_label: MQTT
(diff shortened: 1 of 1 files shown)


docs/software/integrations/mqtt/index.mdx
```diff
@@ -1,8 +1,9 @@
 ---
 id: mqtt
-title: MQTT
+title: MQTT | MQTT Integrations Overview
```
tekstrand, line 3: Do you feel it's necessary to repeat `MQTT` twice here? Wondering if we can shorten to `MQTT | Integration Overview`

tekstrand (approved):
LGTM, one suggestion on repetitive language but I am 👍 either way

PR #1094, opened 2024-03-08
Update web-flasher.mdx

plughie:
The page layout puts a bright green box at the bottom of the page. This draws attention away from the first required step, to flash the firmware. I'm adding text to indicate the order of operations.


tekstrand:
I was going to recommend similar to @rcarteraz here. I might go a bit further and say that we should remove the button and instead use the text `proceed with the initial configuration` as the link **or** perhaps remove that second section altogether and add step 4 that says `Now you may proceed with the configuration...`

tekstrand:
Fair point. I'll withdraw my recommendation here on that and save it for another time :)

PR #1102, opened 2024-03-09
Add footnote for multiple local web clients

### LINKED ISSUE #1097: Updating "Web Client Overview" to call out meshtastic-2.local etc
**Describe what section of the docs is outdated or otherwise in need of change**
"Served directly from an ESP32 based node via [meshtastic.local](http://meshtastic.local/) or the device's IP Address." should be updated e.g. by a foot note

**Suggested Changes**
when multiple Meshtastic nodes are in the same IP network, they can be reached via meshtastic-2.local meshtastic-3.local etc.
Currently this seems to be not documented anywhere and constantly leads to the miss information that the nodes then needs to be reached via their IP address.

rcarteraz:
### What

This PR adds a footnote to the first item in the web client overview regarding the web client being served local from an esp32 device to address when multiple devices are on a single network.

### Why

It's not documented anywhere that multiple devices on the same local area network will adjust the local url as necessary so they can be individually reached. This closes #1097

### DIFF
--- a/docs/getting-started/flashing-firmware/esp32/web-flasher.mdx
+++ b/docs/getting-started/flashing-firmware/esp32/web-flasher.mdx
@@ -1,25 +1,26 @@
 ---
 id: web-flasher
-title: Using Meshtastic Web Flasher
+title: Meshtastic Web Flasher
 sidebar_label: Web Flasher (recommended)
 sidebar_position: 1
 ---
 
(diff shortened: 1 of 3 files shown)


tekstrand (approved):

pdxlocations:
I dig the footnotes!

pdxlocations (approved):

PR #1103, opened 2024-03-10
Add `position_precision` to channels.mdx

### LINKED ISSUE #1058: Document position_precision setting
The new per-channel postion_precision setting needs to be added to docs. Support is soon to land in the mobile and Python clients. In python, it's set via 
```meshtastic --ch-set module_settings.position_precision 13 --port /dev/ttyACM1 --ch-index 0```

The value of 0 means that location data is never sent over the given channel. Value 32 means it is set with full precision. The values in-between indicate the number of bits of precision to be sent. Useful values are roughly 11-16, where 11 indicates a large-ish region of plus or minus 11 Kilometers, and 16 is accurate to within a few doors down, or plus-or-minus 350 meters.

rcarteraz:
### What
Added docs for new `position_precision` channel module setting to control location precision.

### Why
It's important for new features to be documented so users are aware. This closes #1058

### DIFF
--- a/docs/configuration/radio/bluetooth.mdx
+++ b/docs/configuration/radio/bluetooth.mdx
@@ -38,6 +38,10 @@ The default pairing mode will be determined based on whether the device has or d
 
 If your pairing mode is set to `FIXED_PIN`, the default value is **123456**. For all other pairing modes, this number is ignored. A custom integer (6 digits) can be set via the Bluetooth config options.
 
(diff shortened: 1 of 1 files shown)


jp-bennett:
Looks good!

jp-bennett (approved):

pdxlocations (approved):

PR #1104, opened 2024-03-10
Update Bluetooth Fixed_PIN

rcarteraz:
### What

Adds a warning about using default fixed PIN code and recommending changing it. 

### Why 

Using a known default code is a security risk


tropho23 (approved):
Looks great, thank you for adding this.

pdxlocations (approved):

PR #1114, opened 2024-03-11
Correct Python MQTT code

ghost:
paho-mqtt version 2.0 is quite different to the old version 1, so the code needs updating to support the new version


pdxlocations:
I don't have a device at work today to test the code, but the changes look consistent for Paho 2.0
Was the CLA signed and the un-signed?

pdxlocations (approved):

PR #1115, opened 2024-03-11
Linux Native Hardware Support

### LINKED ISSUE #891: Add Linux Native Support
Raspberry Pi support needs to be added to Docs. 

@jp-bennett Has provided a brain dump of information to use as reference for creating the documentation. We'll give a few other devs that participated in development time to add their notes. Doc can be found here: https://docs.google.com/document/d/17LW4ExkG6xVFnusVBMaDMA08kmZfyHiPbi8vjLpmLAs/edit?usp=sharing

rcarteraz:
### What

This PR introduces documentation for setting up Meshtastic on Linux-native devices. It covers the prerequisites, hardware compatibility, installation, configuration for hardware interfaces, and advanced setup, including web server support and troubleshooting.

### Why

To ensure our documentation remains current and comprehensive, making it easier for users to access information about Meshtastic. Closes #891

### DIFF
--- a/docs/community/local-groups.mdx
+++ b/docs/community/local-groups.mdx
@@ -41,7 +41,7 @@ us on [Discord](https://discord.com/invite/ktMAKGBnBs) to add your group.
 - [Hawaii Meshnet](https://www.hawaiimesh.net/)
 
 ### Illinois
-- [Chicagoland Meshtastic](https://discord.com/invite/DUejYDPupT)
+- [Chicagoland Meshtastic](https://chicagolandmesh.org/)
 
 ### Massachusetts
 - [Boston Meshnet](https://github.com/Darachnid/Boston-Meshnet)
@@ -51,6 +51,9 @@ us on [Discord](https://discord.com/invite/ktMAKGBnBs) to add your group.
 
 ### Texas
 - [Austin Mesh](https://austinmesh.org/)
+
+## Poland
+- [Meshtastic Poland Matrix Space](https://matrix.to/#/#meshtasticpl:matrix.org)
 
 ## United Kingdon
 - [UK Meshtastic Kent / South East](https://www.facebook.com/groups/ukmeshtastickent/)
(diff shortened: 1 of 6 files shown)


caveman99 (approved):

jp-bennett:
I helped someone with this just the other day, and it was not clear that when uncommenting lines to enable a radio, they needed to uncomment the module line, as well. So maybe something like:

To enable a LoRa radio connected to your deivce, uncomment the appropriate lines in the config file, including the Module line. As an example, the Waveshare SX126X module would have a Lora section that looks like this:

Lora:
  Module: sx1262  # Waveshare SX126X XXXM
  DIO2_AS_RF_SWITCH: true
  CS: 21
  IRQ: 16
  Busy: 20
  Reset: 18

madeofstown:
Just wanted to request a simple change.

But first, thanks, I've successfully got the Linux Native version of meshtasticd running on an RPi 4b. 

The documents you added say that the PiZero is supported but after 2 days of banging my head against a wall I can't get meshtasticd to build into a native ARM32 binary. The PiZero should probably come off that list of supported devices unless Meshtastic provides a path to a 32bit binary of meshtasticd.

jp-bennett:
> I can't get meshtasticd to build into a native ARM32 binary.

I'd like to add the 32 bit .deb files eventually. For now, the easiest way to compile it is to install the 32 bit OS on a Pi4 or Pi5, build there, and copy the output file over.

You definitely have a valid point that running on the original Pi 0 is non-trivial, and that might warrant a note in the docs.

madeofstown:
Coming back to update. The pi zero does work with Meshtasticd! The build instructions were missing a dependency (`libbluetooth-dev`). Once that dependency has been installed the build will complete!

madeofstown:
Also, it would be nice for some more clarity in the GPS configuration section. It mentions that the pi5 requires certain settings to enable UART connectivity. However, it doesn't say what the correct settings for other pi models are. Can people assume that they are the same as for the pi5 or are these instructions ONLY for the pi5. I was unable to get the GPS working with my Pi4 using those settings as well as some other settings I tried.

jp-bennett:
Based on @madeofstown points, we should add "libbluetooth-dev" as a build dep.

The config.txt GPS line is only for the Pi5.

The gps device for Pi models less than 5 is "/dev/ttyS0"

With those three changes, I say ship it.

jp-bennett:
> The config.txt GPS line is only for the Pi5

Apparently we do need enable_uart=1 on earlier Pi models

jp-bennett (approved):

jp-bennett:
@rcarteraz I love it, let's push it

PR #1119, opened 2024-03-13
Add descriptions to all configuration documents

tekstrand:
## What
Adds descriptions to all pages under Configuration

## Why
- Makes preview links in Discord look better easier to understand
- SEO results look better and should rank higher


tekstrand:
Nope, I think it's just whatever I came up with in between doing other things. We can standardize but it wouldn't make a difference for SEO one way or another AFAIK. I did this way more as a usability thing than an SEO thing.

tekstrand:
I must've missed that PR! I am going to open up more of these over time to get descriptions on all the pages. I think I did "Getting Started" a few weeks back.

PR #1121, opened 2024-03-14
Update MQTT docs for the topic containing 'e'

ianmcorvidae:
Happened to look at this page and noticed it still listed the `/c/` topic structure. I figured we should probably update it, and I added a note about pre-2.3.0 versions as well.


garthvh (approved):
Looks good, thanks!

PR #1127, opened 2024-03-15
Update screens.mdx

nagumii:
added "from left to right" to preferred pin order.


nagumii:
Only if you're used to left to right languages. I don't consider it high
priority, but rtl languages exist, and there are many speakers.

Is there a convention I'm not familiar with when describing pinout
directions?

nagumii:
How about "(as shown below)"?

nagumii:
Honestly, it's just that it confused me for a moment - because I tend to
think in rtl half the time. But you might be right. Feel free to just close
it.

PR #1133, opened 2024-03-17
Note how rebroadcast mode impacts mqtt

tekstrand:
## What
Adds footnotes to both Configuration pages for both Device and MQTT to explain that role or rebroadcast mode will impact MQTT as well.

## Screenshots
Screenshot 2024-03-17 at 4 35 31 PM
Screenshot 2024-03-17 at 4 35 25 PM
Screenshot 2024-03-17 at 4 24 50 PM
Screenshot 2024-03-17 at 4 24 42 PM
Screenshot 2024-03-17 at 4 24 37 PM

### DIFF
--- a/docs/configuration/module/mqtt.mdx
+++ b/docs/configuration/module/mqtt.mdx
@@ -19,7 +19,7 @@ The MQTT module config options are: Enabled, Server Address, Username, Password,
 
 ### Enabled
 
-Enables the MQTT module.
+Enables the MQTT module.[^1]
 
 ### Server Address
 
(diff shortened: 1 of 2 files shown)


Nestpebble:
Maybe it should be on the column heading (Retransmission) rather than just `client_mute`? It also affects `sensor` and `tracker`, for example.

pdxlocations:
> Maybe it should be on the column heading (Retransmission) rather than just `client_mute`? It also affects `sensor` and `tracker`, for example.

I agree with this line of thinking. I'd move the footnote into the main body in the Role Comparison section:
 
Role Comparison
This table shows the default values after selecting a preset. As always, individual settings can be adjusted after choosing a preset. Roles that do not rebroadcast received packets may not be suitable for MQTT gateways.

PR #1159, opened 2024-03-27
Update building.mdx

ch00f:
Added a step that I found non-obvious when trying to to build for custom hardware.


GUVWAF:
This step is unnecessary. If you create the new environment, you should already be able to select it to build and upload.

PR #1160, opened 2024-03-27
Update Tabs CSS

rcarteraz:
Added CSS to change look of tabs. Gives inactive tabs the inactive color to better display to users that they can be clicked. Also makes them line up on the left the same as the content.

image

Will close #650 in favor of this.

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -250,6 +250,8 @@ td {
   border-bottom-color: var(--ifm-tabs-color-inactive-border);
   border-radius: 0;
   margin-right: 4px;
+  padding-left: 12px;
+  padding-right: 12px;
 }
 
 .tabs__item--active {


src/css/custom.css
```diff
@@ -241,3 +241,24 @@ td {
...
   }
 }
+
+.tabs {
+  padding-left: 0;
+}
+
+.tabs__item {
+  border-bottom-color: var(--ifm-tabs-color-inactive-border);
+  border-top-left-radius: 0;
```
tekstrand, line 251: You can simplify these 4 lines into 1 line by using the generic `border-radius`

`border-radius: 0`

src/css/custom.css
```diff
@@ -241,3 +241,24 @@ td {
...
+  border-bottom-left-radius: 0;
+  border-bottom-right-radius: 0;
+  margin-right: 4px;
+  }
+  .tabs__item--active {
+  border-bottom-color: var(--ifm-tabs-color-active-border);
+  }
+  .tabs__item:hover {
+  background-color: var(--ifm-hover-overlay);
+  border-top-left-radius: 8px;
```
tekstrand, line 262: Are you sure you want border radius on hover? Seems abnormal to me to change the radius on something when a user hovers. If so, you can simplify this as well by doing

`border-radius: 8px 8px 0 0;` (args go clockwise starting w/ top left)

tekstrand (changes requested):
My main blocker is actually cleaning up the indentation, other stuff is up to you 🍺 

This looks so much better!

src/css/custom.css
tekstrand: I think just keeping `border-radius: 0` and not changing it on hover is best.

src/css/custom.css
```diff
@@ -241,3 +241,24 @@ td {
     display: block !important;
   }
 }
+
+.tabs {
+  padding-left: 0;
+}
+
+.tabs__item {
```
tekstrand, line 249: Okay one more thing I noticed on this - the left/right padding is not the same, for some reason. We should explicitly set it.
```
padding-left: 12px;
padding-right: 12px;
```

tekstrand (approved):

PR #1163, opened 2024-03-29
Update Wireless Paper

rcarteraz:
### What 

Splits out Wireless Paper into "Wireless Paper V1.0" and "Wireless Paper V1.1". As well as other minor changes for consistency. 

### Why
There's two variants now.


todd-herbert:
There is also a 433MHz option offered, although as best I can tell, the four supposed frequency options are really only two distinct models, a high and a low frequency. The duplicate models are shipped in the same containers with same labelling, antennas have the same characteristics as measured by NanoNVA.

pdxlocations (approved):

PR #1169, opened 2024-04-03
Update contributing.mdx

mc-hamster:


tekstrand (approved):
TIL how to do do alt/title text with markdown.

Only though I had here was if we should host this file on our server in case they remove it from theirs.

mc-hamster:
merging with @tekstrand 's review.

PR #1181, opened 2024-04-09

mzac:
Add link to waveshare lora hat


mzac:
> > Add link to waveshare lora hat
> 
> I'm currently working on an update that will add links but it's not ready yet. Also, the preference will be to link to resources/wiki as listings can change based on stock which requires a lot of maintenance.
> 
> I'm going to close this for now as I'll have an update shortly.

Thanks!  I was finding it hard to find the correct board which is why I thought it would be good to have a link to it, looking foward to seeing your update!

PR #1187, opened 2024-04-15
Update local-groups.mdx

viktormoe:
Added some groups in Sweden + corrected a grammatical error.


pdxlocations:
Thanks for catching that error!  Are you involved with these Sweden groups?  I'm not sure there is a policy in place but I believe it's preferred to have groups add themselves or at least ask them if they want to be added. @rcarteraz ?

viktormoe:
Ok, I understand. I'm waiting to join one of the FB groups but the rest I just found. So maybe its better to scratch this commit, then its only the spell error that needs changing. And I can mention using the mesht trademark in the group.

Issue #1176, opened 2024-04-05
T-Beam Supreme buttons documentation


jgspratt:
https://meshtastic.org/docs/hardware/devices/tbeam/buttons/

Ref https://github.com/meshtastic/firmware/issues/3550

The description of the buttons seems wrong.  I'm still figuring it out, but **the left button appears to be the user button**, not the middle button.  Pressing the left button toggles through the information, while short pressing the middle button does nothing.

I only have T-Beam Supreme units set up so far, so it could be that they differ from T-Beam (non-Supremes).  Or it could be that the firmware has a bug.

Please double check the orientation for the power and reset buttons also.

CyborgySquirrel:
I have the same problem - the button functionality is compromised. The left more button (with screen on the right) scrolls through the different screens and shuts the device down if held long enough. The middle button does nothing (even options to double click to force send position and triple click to turn off/on GPS) doesn’t work. Button three does resent the device.
Would it please be possible to look into this for us? Not being able to interact with the little screen is frustrating.

PR #1190, opened 2024-04-15
Fix docs/hardware/devices/tbeam/buttons.mdx for left vs middle button

jgspratt:
Ref https://github.com/meshtastic/meshtastic/issues/1176

I have two T-Beam Supreme's running 2.3.2 and they both operate with the left button being the user button and the middle button being the power button.  I could be wrong here, but I think this MR corrects a mistake in the documentation.


jgspratt:
Should the Supreme have a separate section? What do you suggest?

pdxlocations:
> Should the Supreme have a separate section? What do you suggest?

I would suggest separate sections for each version if they are different.  I don't have a supreme, so I unfortunately, I can't verify the accuracy of this PR.

PR #1194, opened 2024-04-17
Update usage.mdx

djripcord:
--dest command does not require device ID to be in quotes. In fact, the command fails to match the ID when a user enters it this way.

Modified usage page to reflect this syntax.


GUVWAF:
This depends on the OS, on Linux it doesn't work without. It probably does work with double quotes ("") for you.

pdxlocations:
There is a callout about this on the [remote admin page](https://meshtastic.org/docs/configuration/remote-admin/#remotely-administering-your-node), but I don't think there's a mention of it on the CLI usage page.  @djripcord would you want to add a callout for this reference as well?

pdxlocations:
closing as this change is not correct for all OS's

PR #1196, opened 2024-04-18
Update T-Echo button info

todd-herbert:
Was gently prodded in this direction from https://github.com/meshtastic/firmware/issues/3649#issuecomment-2060506071

Open to any changes / rewrites. Just need to get the updated info out there.


docs/hardware/devices/techo/buttons.mdx
```diff
@@ -1,4 +1,4 @@
----
+    263452---
```
pdxlocations, line 1: what is this number for?

docs/hardware/devices/techo/buttons.mdx
```diff
@@ -7,14 +7,22 @@ sidebar_position: 1
...
-  - **Single press:** Changes the information page displayed on the device's screen.
-  - **Double press:** Turns the screen backlight on/off and sends an adhoc ping of the device's position to the network.
-  - **Long press:** Signals the device to shutdown after 5 seconds.
+  - **Single press:** Power-on / reboot.
+  - **Double press:** Enter bootloader mode, for firmware update.
+- **Program Button (Button 2):**
+  - **Single press:** 
+    - Display next page of information.
+    - *(at screensaver)* Wake the display.
+    - *(when off)* Enter bootloader mode, for firmware update.
```
pdxlocations, line 22: Is this mode any different to a double-press of the top button?

docs/hardware/devices/techo/buttons.mdx
todd-herbert: Actually had no idea! Must have been an accidental keypress I didn't notice

docs/hardware/devices/techo/buttons.mdx
todd-herbert: Exactly the same mode, just reached by a keypress. More of a nuisance than a feature. I'm not sure at what point this behavior changed, but my T-Echo has always done this. I confirmed the behavior yesterday with another user.

Is there a different way you would suggest structuring the info?

pdxlocations: Just curious, I haven't dusted off the ol' T-echo in a while.

docs/hardware/devices/techo/buttons.mdx
```diff
@@ -7,14 +7,22 @@ sidebar_position: 1
...
+  - **Double press:** Send an "adhoc ping": announce device to network.
+  - ** 3x press:** Enable / disable GPS.
+  - ** 4x press:** Enable / disable display's backlight.
+  - **Hold:** Shutdown.
 
 ![TechoButtons](/img/hardware/t-echo-lilygo.webp)
```
pdxlocations, line 28: This image is now incorrect. IMO It should be fixed or removed since it contradicts the new content.

todd-herbert: I believe that image included in this pull-request is an updated version, but it would be good to confirm that it has been included correctly and is visible on your end too.

More than happy to remove it entirely also, if you feel it is no longer necessary.

pdxlocations: Nice!  I'm just unable to preview it.

pdxlocations:
@rcarteraz or @tekstrand could you deploy a Vercel preview?

pdxlocations (approved):

PR #1205, opened 2024-04-22
To add a paragraph under "Tips" with a New pdf file to show Meshtastic Device Screen Info Information

blueskiesMesh:
- I created the pdf file to help users understand the most common info they see on the Meshtastic Node Oled or eInk screen.
  They can download the pdf and print a copy as a quick reference 
- I added a paragraph under "Tips" because I am not sure where else it best to place this pdf file.


pdxlocations:
This is great! 

My thoughts:
1. I'm a little concerned that this document will become outdated and only @blueskiesMesh will be able to update this pdf.  Would it make more sense if this information was created in markdown on it's own docs page?  Perhaps under Hardware?  The download is nice though.  Could the source files be included in the repo for anyone to modify?
2. We typically place images in /static/img/

tekstrand (changes requested):
I feel strongly enough about not adding PDFs that I am marking this as "Request changes" to block merging. @pdxlocations feedback needs to be implemented

This is awesome content, by the way, @blueskiesMesh! Thank you for the great work, let's just get it into a format that matches the rest of the docs so that it can be maintained.

blueskiesMesh:
hi, the original document is in Adobe Indesign . I be glad to share the source file in Adobe InDesign format. Any suggestion on what format will work ?

tekstrand:
@blueskiesMesh the best format is going to be the same that the rest of the documentation for Meshtastic already uses which is Markdown. An individual PDF can then be generated from that page on the website, and it will also be included in the full documentation PDF already available for download on our Downloads page https://meshtastic.org/downloads/

pdxlocations:
I'd love to have this content integrated in the docs.  Questions about the display are common enough.  I don't think it would be possible to create this using markdown without relying heavily on rendered images.  Adding the source files to the repo would give the community access to edit the images if/when changes are made but InDesign isn't something everyone has access to.  Could these images be recreated in something free like google slides or inkscape?  Just a thought.

PR #1212, opened 2024-04-23
Add German language telegram groups

caveman99:


caveman99:
I am aware of that and we are in the process of changing that. For Regio I have no immediate idea for a catchy name that best describes the function without violating the letter of the grant. The group sprung out of necessity not to clutter the main group.

caveman99:
Icons have been changed.

pdxlocations:
> Icons have been changed.

The D-A-CH looks superb!  The other group though seems to me to still be in conflict with the trademark guidelines for both naming and logo use.  What do you think?

caveman99:
> The D-A-CH looks superb! The other group though seems to me to still be in conflict with the trademark guidelines for both naming and logo use. What do you think?

It's not easy. The current changes have caused a bit of a stir already. Main Group is D-A-CH. The regio group is for local (or state) coordination of network buildup. It is linked after signing in to the main group though, so we might as well remove it from the listing.

pdxlocations (approved):

PR #1214, opened 2024-04-23
remove unused Mapbox reference from Meshtastic privacy policy

andrekir:
Since 2.0 Mapbox is no longer used and this reference can be removed.


pdxlocations (approved):

garthvh (approved):
Approved

PR #1226, opened 2024-04-28
shutdown after losing power - check power detection status note

fifieldt:
As described in meshtastic/firmware#3743 , some users were discovering that their devices unexpectedly powered down after setting "Shutdown on battery delay", despite their devices being connected to USB.

This was due to the devices not correctly detecting that they were connected to USB. This can be resolved by altering the ADC_MULTIPLIER per existing instructions.

This commit adds a note to encourage users to check the power status before enabling "Shutdown on battery delay".

fixes meshtastic/firmware#3743

### DIFF
--- a/docs/configuration/radio/power.mdx
+++ b/docs/configuration/radio/power.mdx
@@ -30,6 +30,9 @@ When activated, this feature disables Bluetooth, Serial, WiFi, and the device's 
 
 Automatically shut down a device after a defined time period if power is lost.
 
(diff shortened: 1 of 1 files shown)


geeksville:
This pull request has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/question-re-configuration-power-shutdown-on-battery-delay/12241/10

pdxlocations (approved):

PR #1227, opened 2024-04-28
Update local-groups.mdx

bitnapper:
Adding new German Matrix space


pdxlocations (changes requested):
Thanks for your contribution however it appears the group is not compliant with the allowed usage of "Meshtastic" and M logo for social media sites:

bitnapper:
Ok, I'll delete it then.

PR #1228, opened 2024-04-29
correct positioning

rcarteraz:
Updated sidebar position


pdxlocations:
Canary is also listed first on:  https://meshtastic.org/docs/hardware/devices/
I'm curious what the correct order would be?

pdxlocations (approved):

PR #1203, opened 2024-04-22
Update _lora-regions.mdx: add Brazil, Argentina to ANZ

AndreiRegiani:
image

Source: https://www.thethingsnetwork.org/docs/lorawan/frequencies-by-country/

* `AU915-928`: Brazil, Argentina

PR #1230, opened 2024-04-29
create `LoRa region by Country` table

andrekir:
follow up on https://github.com/meshtastic/meshtastic/pull/1203#issuecomment-2081427391

let's see how this looks

### DIFF
--- a/docs/configuration/region-by-coutry.mdx
+++ b/docs/configuration/region-by-coutry.mdx
@@ -1,0 +1,147 @@
+---
+id: region-by-country
+title: LoRa Region by Country
+description: List countries associated to their respective LoRa Region settings.
+sidebar_label: LoRa Region by Country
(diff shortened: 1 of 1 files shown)


pdxlocations:
This is great!  Would it be helpful to include specifics such as frequency range and duty cycles?

andrekir:
> This is great! Would it be helpful to include specifics such as frequency range, spacing, power limit and duty cycles found in: https://github.com/meshtastic/meshtastic/blob/master/src/components/tools/FrequencyCalculator.tsx

this table is for users just getting started with Meshtastic, so I think adding all that info is a bit overwhelming.

but I do agree we are missing detailed region info in the docs, perhaps that should go here https://meshtastic.org/docs/configuration/radio/lora/#region and we can point to it.

pdxlocations (approved):

PR #1232, opened 2024-04-30
Add two entries to FAQ regarding firmware Update faq.mdx

mkgin:
Add two entries to FAQ regarding firmware.
* Difference between Beta amd Alpha
* When to Update firmware

### DIFF
--- a/docs/about/faq.mdx
+++ b/docs/about/faq.mdx
@@ -25,6 +25,14 @@ export const Faq = {
       title: "Can I update my node's firmware over the mesh?",
       content: `No, Meshtastic does not support OTA updates over LoRa.  Please visit [Flash Firmware](/docs/getting-started/flashing-firmware/) for update options.`,
     },
+    {
+      title: "Which version of firmware should I use?",
+      content: `Currently, there are two types of Meshtastic firmware releases, "Alpha" and "Beta". Firmware releases are labeled with the version number, commit hash and the type. Installing the latest beta release recommended for most cases. Alpha release are geared towards users interested in testing the new features, fixes or enhancements with the understanding there are additional risks and potential for instability compared to running a beta release.`,
+    },
+    {
+      title: "How often should I update firmware?",
+      content: `Firmware should be kept up to date. While it is not necessary to update whenever a new release comes out it's a good idea to follow the Discord Server and Forums linked in the [Introduction](/docs/introduction/#contributors) where developers may advise the community about important updates.  The Meshtastic [firmware releases page](https://github.com/meshtastic/firmware/releases) on Github has a summary of what has changed in each release.`,
+    },
   ],
   "android": [
     {


docs/about/faq.mdx
```diff
@@ -25,6 +25,14 @@ export const Faq = {
       title: "Can I update my node's firmware over the mesh?",
       content: `No, Meshtastic does not support OTA updates over LoRa.  Please visit [Flash Firmware](/docs/getting-started/flashing-firmware/) for update options.`,
     },
+    {
+      title: "Which version of firmware should I use?",
+      content: `Currently, there are two types of Meshtastic firmware releases, "Alpha" and "Beta". Firmware releases are labeled with the version number, commit hash and the type. Installing the latest beta release recommended for most cases. Alpha release are geared towards users interested in testing the new features, fixes or enhancements with the understanding there are additional risks and potential for instability compared to running a beta release.`,
```
pdxlocations, line 30: I'm not sure I know what "type" is referring to in "version number, commit hash and the type"

pdxlocations (commented):
I find these additions to be very helpful!

docs/about/faq.mdx
mkgin: Release type... alpha and beta.  It could be rephrased to be more clear. 

If there was a more formal description of the differences in release types used for firmware or policy,  I'd just link there. (now other places can link here so I try to keep it very general for now.)  maybe someone involved with firmware wants to say something? 

In the web flasher, releases are categorized as Stable ( with version string like 2.3.7.abde123 beta ) and alpha. 

I think the following could be improved:
- type ... rephrase, maybe a better word?
- beta ... is it beta or stable? It is both? ... perpetual beta is a nice concept when things are ad hoc.

I'll reply later (probably tomorrow) with some updated text. 

When the text looks good, I'll make a new commit and check with vercel.

turnrye: It's vital that the naming is consistent across the user-facing surfaces. Given that most users would interact through the web flasher, and it says Stable, I think it should say the same thing elsewhere on the website. Consistency.

The tags "alpha" and "beta" represent a pre-release build in most software versioning systems. There are a few things in our ecosystem here that point to these "Stable" releases not being pre-release. Hardware vendors are confident enough in them to ship products with it preinstalled, the number of open GitHub issues is relatively low, and perhaps most importantly it doesn't appear that there are additional builds with the same major.minor.patch under this "beta" tag. So at this point, it appears as though "beta" is actually the general release (and it's marked as such on GitHub releases as well). 

So... this points to Meshtastic having two build _channels_, a **pre-release** build channel of **testing** builds that have "Alpha" in the name, and a **release** build channel of **stable** builds that have "Beta" in the name. I'd suggest ignoring the naming details in user facing documentation, because the versioning scheme isn't important to communicate to our users. So don't mention Alpha or Beta at all, and don't mention commit hash (I think most people who know what that means would immediately recognize the hash anyway).

pdxlocations: The web flasher does list the firmware versions Beta and Alpha, under the headers of Stable and Alpha so I think the naming convention is somewhat explained for the user. 
Screenshot 2024-05-01 at 8 31 51 AM

In regards to this PR, maybe instead of just "type" I would suggest "Beta or Alpha". or "type (Beta or Alpha)"

mkgin: On discord, someone was having problems with something, mentioned running "alpha" with the impression that newer is better. 

The community on discord and forums come from a wider group than a typical github user. Things should be consistent across the platform. Some things should be spelled out, but not with more detail than necessary.

Separate (but connected) future issues:
* **web flasher** is super convenient for alpha and beta users.  Something should be updated there too.
  * link to FAQ
  * correct release terminology
  * latest Stable pre-selected by default
* release info
  * something brief and "official" regarding build channels as described earlier
  * description of how release schedule could be nice (even if it is vague )
  * Something in Development/Firmware documentation or even easier, quick note in the firmware README.md
  
All this for a couple paragraphs :-) but I'd rather get it right without too many iterations.

pdxlocations: I'd say the paragraphs look fine, but perhaps clarify what "type" meant.  It took me a minute to realize what you are referring to.  Regarding the other issues, feel free to open additional feature requests or PR's.  I personally don't feel too strongly one way or another about the terminology.

pdxlocations: re-reading it and it actually makes sense now.  I was super tired the first time I looked at this!

pdxlocations (approved):

PR #1235, opened 2024-05-01
Add Frequency Details to LoRa Regions

pdxlocations:
This PR adds the frequency range, duty cycle and power limits to the region code list.
Values taken from [RadioInterface.cpp](https://github.com/meshtastic/firmware/blob/master/src/mesh/RadioInterface.cpp)

Screenshot 2024-04-30 at 10 40 07 PM


andrekir (approved):
looks good!

PR #1238, opened 2024-05-02
Add BMP085/180 and INA3221 sensors, alternate INA2XX address

ianmcorvidae:
Would appreciate a double-check if someone's got the ability, but I also grabbed these values from the firmware code directly:

https://github.com/meshtastic/firmware/blob/master/src/detect/ScanI2CTwoWire.cpp#L238 for BMP085/180, line 262 has a third INA2XX address, and 273 has the INA3221 details.

I'm assuming that accelerometers and magnetic sensors don't go on this page. At least, if they do, I'm not making anything worse, haha.


pdxlocations:
This is beyond me, @rcarteraz know anything about this?

ianmcorvidae:
realizing I could also ask the firmware folks who added some of these if I've got it right -- looks like @thoherr added the BMP085/180 stuff, @caveman99 looks to have added the third INA2XX address, and @RicInNewMexico is who added the INA3221 support. Do these additions look correct to you three?

caveman99 (approved):

pdxlocations (approved):

PR #1244, opened 2024-05-06
Add Czechia to the list of countries

piit79:


piit79:
Sure. Although the new(ish) "official" country name is Czechia, Czech Republic is also fine and still used in many cases.

caveman99:
I chose the more formal version to not have people confuse it with Chechnya. I know the Austria/Australia problems ... :-)

PR #1254, opened 2024-05-10
Update buttons.mdx

rcarteraz:
Removes mention of "if tap to wake is enabled" that was removed with https://github.com/meshtastic/firmware/pull/3745


todd-herbert:
Thanks for that, sorry.. I should have remembered..

pdxlocations (approved):

PR #1266, opened 2024-05-24
Add a section to show how to create device trackers in HA to show nodes on maps

ianmcorvidae:
It's also sorta-possible to do this using the values from the sensors created by the rest of the docs, but can lead to an odd situation where HA calls the device tracker service twice, one that changes the latitude and one that changes the longitude (or the opposite order), creating a weird "blocky" look. This way only ever pulls it in when both values are defined and is clean enough for me to actually put it in a PR finally.

### DIFF
--- a/docs/software/integrations/mqtt/home-assistant.mdx
+++ b/docs/software/integrations/mqtt/home-assistant.mdx
@@ -176,6 +176,40 @@ sensor:
       {% endif %}
 ```
 
+### Maps & Device Tracker
+
+To make it possible for your nodes to appear in maps within Home Assistant, you will need to set up a [device_tracker](https://www.home-assistant.io/integrations/device_tracker/) entity. Luckily, Home Assistant allows them to be created ad-hoc, simply by calling the `device_tracker.see` service from a script or automation.
+
+To do this, create a new automation and use the three-dots menu to change to the "Edit in YAML" mode, and set it up like the following:
+
+```yaml
+alias: Update Node 1 location
+description: Update Meshtastic node when corresponding MQTT messages are seen.
+trigger:
+  - platform: mqtt
+    topic: msh/2/json/LongFast/!67ea9400
+    payload: "on"
+    value_template: |-
+      {% if value_json.from == ("f0b94b6d" | int(base=16)) and
+            value_json.payload.latitude_i is defined and 
+            value_json.payload.longitude_i is defined %}on{% endif %}
+condition: []
+action:
+  - service: device_tracker.see
+    metadata: {}
+    data:
+      dev_id: node_1
+      gps:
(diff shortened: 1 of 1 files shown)


docs/software/integrations/mqtt/home-assistant.mdx
```diff
@@ -176,6 +176,40 @@ sensor:
...
+alias: Update Node 1 location
+description: Update Meshtastic node when corresponding MQTT messages are seen.
+trigger:
+  - platform: mqtt
+    topic: msh/2/json/LongFast/!67ea9400
+    payload: "on"
+    value_template: |-
+      {% if value_json.from == ("f0b94b6d" | int(base=16)) and
+            value_json.payload.latitude_i is defined and 
+            value_json.payload.longitude_i is defined %}on{% endif %}
```
ianmcorvidae, line 195: The way this works is that the `value_template` output is compared to the `payload` field, so this sets it up to return `on` if it's the correct node & the lat/lon are defined. I'm not sure if this mechanism should be explained in more depth on the page, or just let people google it/ask in discord.

I also have this snippet calculating the integer node ID from the hexadecimal one; unsure if that's better or not (`f0b94b6d` is 4038675309 that we're using elsewhere in this page).

In the future I may put in a much bigger rework of this page that uses an MQTT trigger like this for most of the sensors (template sensors can, these days, also use triggers, so they could be set up to only update on the correct messages, rather than relying on the `else` blocks we currently are, but it'll all be in totally different files and sections of config, so I've been putting off building it for myself/updating docs.

pdxlocations: > I also have this snippet calculating the integer node ID from the hexadecimal one; unsure if that's better or not (`f0b94b6d` is 4038675309 that we're using elsewhere in this page).

 4038675309 is Canadian Jenny  😉

pdxlocations: Would you consider leaving the from as Jenny for now to be consistent with the rest of the page?  Not sure if either is better but I do appreciate a good consistency.

pdxlocations:
I've added my car node as a tracker!  Very nice addition to the docs.

docs/software/integrations/mqtt/home-assistant.mdx
ianmcorvidae: I totally didn't catch that 😂 I'll swap it back. Maybe I'll make a different pull request to show the conversion trick somewhere in the page

pdxlocations (approved):

docs/software/integrations/mqtt/home-assistant.mdx
pdxlocations: That would be cool.  You could also just add a comment in the code block with the option, like I did with the F/C conversion.

Issue #1259, opened 2024-05-17
Documentation Change Request : Linux Native Agent - systemctl and log rotation


Talie5in:
**Describe what section of the docs is outdated or otherwise in need of change**
Linux Native Hardware

**Suggested Changes**
Add After "Persistence" or part of "Persistence".
https://meshtastic.org/docs/hardware/devices/linux-native-hardware/#persistence

----

## Run Meshtastic as a service/daemon for manual compile/builds (**non .deb package releases**)
To Setup the device to start and stop meshtasticd as as service using systemctl you can setup the service unit using the instructions below.

Create the service unit file:

Create a new file in the /etc/systemd/system/ directory with a name like meshtasticd.service.
```sudo nano /etc/systemd/system/meshtasticd.service```
Add the following content to the file:
```
[Unit]
Description=Meshtastic Daemon
After=network.target

[Service]
ExecStart=/usr/sbin/meshtasticd
Restart=always
User=root
Group=root
StandardOutput=file:/var/log/meshtasticd.log
StandardError=file:/var/log/meshtasticd.log
Type=simple

[Install]
WantedBy=multi-user.target
```

Reload systemd to recognize the new service:
```sudo systemctl daemon-reload```

Enable the service to start on boot:
```sudo systemctl enable meshtasticd```

### Starting and Stopping the Service

Start the service:
```sudo systemctl start meshtasticd```

Check the status of the service:
```sudo systemctl status meshtasticd```

Stop the service:
```sudo systemctl stop meshtasticd```

This will give you a detailed view of the service status and any potential errors.

By following these steps, you set up a systemd service for meshtasticd that will start automatically at boot and restart if it crashes. You can manage it using the standard systemctl commands (start, stop, restart, status, etc.).

## Setup Log Rotation
To setup log rotation within the Linux Native Device and do a daily restart of the Meshtastic Daemon you can implement the below.

Create a logrotate configuration file for the meshtasticd log. Create a file named /etc/logrotate.d/meshtasticd:
```sudo nano /etc/logrotate.d/meshtasticd```

Add the following content to the file:
```
/var/log/meshtasticd.log {
    daily
    missingok
    rotate 7
    compress
    delaycompress
    notifempty
    create 0640 root root
    sharedscripts
    postrotate
        systemctl restart meshtasticd > /dev/null 2>&1 || true
    endscript
}
```

### Test log rotation:

Test the logrotate configuration to ensure it works as expected:
[...]

PR #1267, opened 2024-05-25
Updates to Linux Native Devices documentation

Talie5in:
Update of Linux Native Agent documentation to satisfy https://github.com/meshtastic/meshtastic/issues/1259
This should also allow us to close the issue on the firmware https://github.com/meshtastic/firmware/issues/3909 as GPSFan wanted it left open until documentation was updated.

Added commands to setup config.txt and cmdline.txt correctly using mostly raspi-config commands instead of manual edits, but left manual method documentation in tact.

Also expanded on some Advanced Functions (eg, those with manual builds/non .deb releases) to create a service script and logrotate.

### DIFF
--- a/docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
+++ b/docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
@@ -169,7 +169,6 @@ ExecStart=/usr/sbin/meshtasticd
 Restart=always
 User=root
 Group=root
-StandardOutput=file:/var/log/meshtasticd.log
 StandardError=file:/var/log/meshtasticd.log
 Type=simple
 
@@ -208,45 +207,12 @@ This will give you a detailed view of the service status and any potential error
 
 By following these steps, you set up a systemd service for meshtasticd that will start automatically at boot and restart if it crashes. You can manage it using the standard systemctl commands (start, stop, restart, status, etc.).
 
-#### Setup Log Rotation
-To setup log rotation within the Linux Native Device and do a daily restart of the Meshtastic Daemon you can implement the below.
+#### View Logs of Mesthastic
+To view the log output of the meshtasticd service, use the below command to read them out of the system journal.
 
-Create a logrotate configuration file for the meshtasticd log. Create a file named /etc/logrotate.d/meshtasticd:
 ```shell
-sudo nano /etc/logrotate.d/meshtasticd
+journalctl -u meshtasticd -b
 ```
-
-Add the following content to the file:
-```plaintext
-/var/log/meshtasticd.log {
-    daily
-    missingok
-    rotate 7
-    compress
-    delaycompress
-    notifempty
-    create 0640 root root
-    sharedscripts
-    postrotate
-        systemctl restart meshtasticd > /dev/null 2>&1 || true
-    endscript
-}
-```
-
-##### Test log rotation:
-
-Test the logrotate configuration to ensure it works as expected:
-```shell
-sudo logrotate -d /etc/logrotate.d/meshtasticd
-```
-
-If everything looks good, force a rotation to see it in action:
-```shell
-sudo logrotate -f /etc/logrotate.d/meshtasticd
-```
-
-By following these steps, you will have meshtasticd output directed to a log file with automatic log rotation configured, ensuring that your logs are managed and disk space is conserved.
-
 
 #### Installation of drivers
 
--- a/docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
+++ b/docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
@@ -62,7 +62,7 @@ For devices requiring SPI or I2C:
 
 ```shell
 sudo raspi-config nonint set_config_var dtparam=spi on /boot/firmware/config.txt # Enable SPI
-sudo raspi-config nonint set_config_var dtparam=i2c_arm on /boot/firmware/config.txt # Enalbe i2c_arm
+sudo raspi-config nonint set_config_var dtparam=i2c_arm on /boot/firmware/config.txt # Enable i2c_arm
 
 # Ensure dtoverlay=spi0-0cs is set in /boot/firmware/config.txt without altering dtoverlay=vc4-kms-v3d or dtparam=uart0
 sudo sed -i -e '/^\s*#\?\s*dtoverlay\s*=\s*vc4-kms-v3d/! s/^\s*#\?\s*(dtoverlay|dtparam\s*=\s*uart0)\s*=.*/dtoverlay=spi0-0cs/' /boot/firmware/config.txt
--- a/docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
+++ b/docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
@@ -58,7 +58,22 @@ sudo apt install ./meshtasticd_{version}arm64.deb
 
 For devices requiring SPI or I2C:
 
-- SPI support can be enabled in `/boot/config.txt` or `/boot/firmware/config.txt`:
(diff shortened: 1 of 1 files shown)


docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
```diff
@@ -114,13 +136,120 @@ dtoverlay=uart0
...
+[Unit]
+Description=Meshtastic Daemon
+After=network.target
+
+[Service]
+ExecStart=/usr/sbin/meshtasticd
+Restart=always
+User=root
+Group=root
+StandardOutput=file:/var/log/meshtasticd.log
```
jp-bennett, line 172: For now, let's drop the txt log output and log rotation. I'm very tempted to add this to the official .deb, but until that happens, let's keep the instructions in sync.

docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
```diff
@@ -114,13 +136,120 @@ dtoverlay=uart0
...
+Stop the service:
+```shell
+sudo systemctl stop meshtasticd
+```
+
+This will give you a detailed view of the service status and any potential errors.
+
+By following these steps, you set up a systemd service for meshtasticd that will start automatically at boot and restart if it crashes. You can manage it using the standard systemctl commands (start, stop, restart, status, etc.).
+
+#### Setup Log Rotation
```
jp-bennett, line 211: And so here, we could simply inform users that `journalctl -u meshtasticd -b` will fetch the meshtasticd logs since boot, and `journalctl -u meshtasticd -f` will follow the live log output.

jp-bennett (changes requested):

docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
Talie5in: Makes sense, removed StandardOutput= entry from service file.

docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
Talie5in: Makes sense, removed Logrotate script and updated with information on how to view logs from the journal.

Talie5in:
@jp-bennett Made changes as suggested.

jp-bennett (approved):

docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
```diff
@@ -58,7 +58,22 @@ sudo apt install ./meshtasticd_{version}arm64.deb
 
 For devices requiring SPI or I2C:
 
-- SPI support can be enabled in `/boot/config.txt` or `/boot/firmware/config.txt`:
+- This can be done by running the below commands on a Raspberry Pi (2-5)
+
+```shell
+sudo raspi-config nonint set_config_var dtparam=spi on /boot/firmware/config.txt # Enable SPI
+sudo raspi-config nonint set_config_var dtparam=i2c_arm on /boot/firmware/config.txt # Enalbe i2c_arm
```
wheresalice, line 65:
```suggestion
sudo raspi-config nonint set_config_var dtparam=i2c_arm on /boot/firmware/config.txt # Enable i2c_arm
```

docs/hardware/devices/linux-native-hardware/linux-native-hardware.mdx
```diff
@@ -114,13 +136,85 @@ dtoverlay=uart0
...
 
 ### Persistence
 
 - The persistent .proto db files of the portduino version of meshtasticd are stored under: `/root/.portduino/default/prefs/`.
 
 ### Advanced Setup and Troubleshooting
 
+#### Creating and Enabling a Systemctl Script (non .deb installs)
+
+To Setup the device to start and stop meshtasticd as as service using systemctl you can setup the service unit using the instructions below.
```
wheresalice, line 152:
```suggestion
To configure the device to start and stop meshtasticd as as service using systemctl you can setup the service unit using the instructions below.
```

GPSFan:
Does this eliminate the need to disable or mask the getty service that is on /dev/ttyS0?
There is also a lot of text that comes out the serial port during a reboot, this may cause confusion in some GPSs, I haven't seen any so far though.

tekstrand:
Are we sure this should all be under `Hardware` and not `Software`?

jp-bennett:
> Are we sure this should all be under `Hardware` and not `Software`?

We put it under hardware, as it's another hardware target.

wheresalice (approved):

Talie5in:
Reviewed the suggested changes and commited.

Let's try a final review before push.

Talie5in:
> Does this eliminate the need to disable or mask the getty service that is on /dev/ttyS0?

This part in the updated doco covers this.

https://github.com/meshtastic/meshtastic/pull/1267/commits/5c16cd0b381beb1bef87f94057babeab5d2e81ce#diff-a98d35af51676f7add00c680c0e59617ab493fc7834f402f957af28d3eaabde0L122-L125


> There is also a lot of text that comes out the serial port during a reboot, this may cause confusion in some GPSs, I haven't seen any so far though.

I believe this is correctly handled when meshtasticd detects the matching GPS Chip, and starts it's initialization strings to put the GPS into the correct mode...

https://github.com/meshtastic/firmware/blob/master/src/gps/GPS.cpp#L402-L436

GPSFan:
I'll have to reinstall everything on my Pi to get to an unaltered fresh install state, to try this out, but I couldn't get rid of the getty on /dev/ttyS0 without masking the service (disabling didn't survive a reboot). None of the serial configuration commands seemed to do that.
Re the data that comes out the serial port before the GPS Init, if it overwhelms the receive buffer in the GPS chip, data that follows (the config commands) COULD be missed/lost.

tekstrand:
> > Are we sure this should all be under `Hardware` and not `Software`?
> 
> We put it under hardware, as it's another hardware target.

As of today, the other hardware pages do not have content similar to this. Under Software > Linux we have existing Docker  documentation. In my view we need to decide one place and stick with it meaning this needs to go under software, or the docker instructions need to come here.

@pdxlocations @rcarteraz what are y'alls thoughts on this?

jp-bennett:
> As of today, the other hardware pages do not have content similar to this. Under Software > Linux we have existing Docker documentation. In my view we need to decide one place and stick with it meaning this needs to go under software, or the docker instructions need to come here.

None of the other hardware pages have this sort of content, because we ship the entire firmware for those targets, while here we just the daemon that requires other system changes to work. The Docker instructions are a bit different as they only apply to running the native target as a simulator (for now). 

I agree that it's weird that we split the information across the two pages, but it does serve a purpose. Perhaps in the future we can re-organize which bits are where.

Talie5in:
> but I couldn't get rid of the getty on /dev/ttyS0 without masking the service (disabling didn't survive a reboot). None of the serial configuration commands seemed to do that.

This specific command here..

```
sudo raspi-config nonint do_serial_cons 1
```

Is running this (which i've pinched out of the raspi-config script) 
```Remove any occurrence of 'console=ttyAMA0,<baudrate>' or 'console=serial0,<baudrate>'```
https://github.com/meshtastic/meshtastic/issues/1259#issuecomment-2119130453

Removes the serial port and baud rate from cmdline.txt which is what the getty service gets created from on boot (clearly a script elsewhere I havent found but no-really-a-problem.)

Once the serial string is removed from cmdline.txt then the service doesnt autocreate and start on boot and tie up the serial port - from what I can see it also means the kernal-boot process doesnt attach to the serial port either.

I've tested this on a rPi3 on fresh Raspbian image as part of writing up the docs.

GPSFan:
@Talie5in That makes sense, I was sure I tried that combo, but maybe not. As to where to put it, choose one and put a link in the other as this really configuration is both hardware and software.

PR #1278, opened 2024-05-30
Update radio-settings.mdx

jwcrawley:
Include the default frequencies for the associated presets set within the app.

### DIFF
--- a/docs/about/overview/radio-settings.mdx
+++ b/docs/about/overview/radio-settings.mdx
@@ -74,16 +74,16 @@ Various data-rate options are available when configuring a frequency slot and ar
 
 We have eight LoRa radio presets. These are the most common settings and have been proven to work well:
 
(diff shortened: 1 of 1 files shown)


tekstrand (changes requested):
Wouldn't these only be accurate for the US? That's what the calculator up above on the page is for https://meshtastic.org/docs/overview/radio-settings/#frequency-slot-calculator

If we'd like to display them all without having to use the calculator, I'd suggest we create a table with all regions or a table for each region, rather than adding it to this table

pdxlocations:
@tekstrand is correct, this PR would only apply to the US region setting and could be misleading.
There is a default LongFast frequency slot list here: https://meshtastic.org/docs/configuration/tips/#default-primary-frequency-slots-by-region
A frequency rage table here:
https://meshtastic.org/docs/configuration/radio/lora/

For preset frequencies by country the calculator should be used.

PR #1279, opened 2024-06-01
Update local-groups.mdx

elranr:
Added Israel Meshtastic club telegram group


garthvh:
Is meshtastic legal is Israel? Previous research indicates no..

pdxlocations (approved):
I hope you don't mind I edited your PR to correct the order and header level

tekstrand:
> Is meshtastic legal is Israel? Previous research indicates no..

Do you recall why it was be illegal? I believe the 915mhz ISM band is allowed there.

garthvh:
Looks like Lora would require a 1% duty cycle

caveman99:
> Looks like Lora would require a 1% duty cycle

Yea, but couldn't a user group also be advocating to make it legal? Having a local group is not illegal per se, so i think we can accept the listing.

tekstrand (approved):

PR #1286, opened 2024-06-05
Update trademark-grants.mdx

IhorBats:
Added my Etsy shop information. 
Meshtastic listings are in delisted state(not visible) as they contain M logo and I need permission to use it.


IhorBats:
Thanks, I will use M-PWRD on my products going forward. Still will be sponsoring you and donating :)

PR #1300, opened 2024-06-19
Update local-groups.mdx

Kavitate:
Added midwest mesh group for qualifying states.


pdxlocations:
I'm going to hold off approving this unless an admin approves the use of the Meshtastic Trademark on your Discord Server.  Alternatively you may wish to create a unique brand for your club.

image

Kavitate:
I can change it to "Midwest Mesh" and change the logo if needed.

Kavitate:
@pdxlocations I've updated the group name and server logo.

pdxlocations:
Thanks!  OK, now I'm not too keen on linking 6 times from different states.  I don't believe that's how the page was intended to be used.  If you're not associated with particular state, my opinion would be to pick one, or just list it as a group under United States, NOT affiliated with a state.

See how PDXMesh mentions WA in the description and/or how UK has a "North East England" link

Kavitate:
@pdxlocations I wasn't sure how they would want them listed since the U.S is broken down by states and no one has done a region yet. Anyway, I've created the "Midwest Region" under the U.S and added the group in there.

pdxlocations:
I think we can leave out the Midwest Region header. If someday there are more Midwest clubs we can add the header back in.  Open to other ideas of course.

Kavitate:
Removed the Midwest Region header.

pdxlocations (approved):

PR #1310, opened 2024-06-23
Modified and expanded Peripherals

noon92:
Added vibramotor and 3rd party passive buzzer and noted that the rak buzzer, as currently implemented, is extremely quiet.

### DIFF
--- a/docs/hardware/devices/rak/peripherals.mdx
+++ b/docs/hardware/devices/rak/peripherals.mdx
@@ -16,6 +16,7 @@ values={[
 {label: 'User Button', value: 'User Button'},
 {label: 'GPS Module', value: 'GPS'},
 {label: 'Buzzer', value: 'Buzzer'},
+{label: 'Vibramotor', value: 'Vibramotor'},
 {label: 'I/O Module', value: 'IO'},
 {label: 'Environmental Sensors', value:'Sensors'},
 {label: 'RTC Module', value:'RTC'}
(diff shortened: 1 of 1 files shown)


docs/hardware/devices/rak/peripherals.mdx
```diff
@@ -89,16 +90,39 @@ The RAK1910 is supported on the following base boards & slots:
 </TabItem>
 <TabItem value="Buzzer">
 
-## Buzzer
```
tekstrand, line 92: This heading should remain, since it is the heading for the tab, and then there should be a new third level heading under it.

docs/hardware/devices/rak/peripherals.mdx
```diff
@@ -16,6 +16,7 @@ values={[
 {label: 'User Button', value: 'User Button'},
 {label: 'GPS Module', value: 'GPS'},
 {label: 'Buzzer', value: 'Buzzer'},
+{label: 'Vibramotor', value: 'Vibramotor'},
```
tekstrand, line 19: This tab will not work because it does not match the `value` of your `TabItem` lower in the file. I would recommend updating the value to match the title of the section. Might I recommend `Vibration` as the title?

docs/hardware/devices/rak/peripherals.mdx
```diff
@@ -89,16 +90,39 @@ The RAK1910 is supported on the following base boards & slots:
...
+### Resources
+- Purchase Links:
+  - International
+    - Amazon: [Buzzer](https://www.amazon.com/dp/B018I1WBNQ)
+    - Amazon: [Transistor](https://www.amazon.com/dp/B07YN7KQQB)
+
+</TabItem>
+<TabItem value="Vibration">
+
+## Third Party Buzzer
```
tekstrand, line 116: Update this to match the tab title `Vibration`

tekstrand:
No fundamental concerns. Fix up these few things and you should be good to go.

noon92:
Sorry for the delay - I think that's everything.

docs/hardware/devices/rak/peripherals.mdx
tekstrand: @noon92 sorry I am being picky here but I'd really like us to make sure headings match tabs. Can you please update this to still be buzzer and add a new heading below for RAK? Thank you!

noon92:
@tekstrand Is this what you meant? Sorry for the delay.

tekstrand (changes requested):
No. You need to have a `Buzzer` heading above `RAK Buzzer`. Headings should always be sequential in level. 

This is what you need

```mdx
## Buzzer
### RAK Buzzer
```

noon92:
Yep, good point @tekstrand. I think that's handled now.

tekstrand (approved):
👍 thanks for getting this fixed up and adding great info!

noon92:
Is this still unmerged? @rcarteraz?

tekstrand:
@rcarteraz cut it out. You pressed "Request changes" so you take responsibility of following up or dismissing it, don't put this on me.

tekstrand:
> That type of reply is completely uncalled for

Your passive aggressive behavior here _does_ call for a response. My comment is **neither** rude nor uncalled for.

There is no issue with a second reviewer requesting changes and then approving once their requested changes are approved. However, until all reviewers have resolved their request changes the PR cannot be merged. Each reviewer is responsible for their approvals and requests. No one else.

tekstrand:
> Honestly, every time someone replies with something you don’t like you immediately label it as “passive aggressive”

I am open to hearing feedback on any specific instances you would like to share as I am not aware of this pattern to date.

PR #1318, opened 2024-06-30
Clarify the necessity and configuration of the "mqtt" Meshtastic channel

zyphlar:
My MQTT wasn't working at all (nothing published to msh/US/2/# except msh/US/2/stat/!myradio = offline for a week) until on advice from Discord I added a new Meshtastic channel to my radio configuration called "mqtt" and then everything started working. They claim that this paragraph in the documentation makes that requirement clear, but:

1. This paragraph is only about JSON downlink to send MQTT over LoRa, which is odd because MQTT uplink didn't work for me either without these instructions, and anyway also:
2. The word "channel" by itself isn't a full set of instructions for creating a new Meshtastic channel on the appropriate node. The rest of the paragraph talks about the /mqtt/ topic, which is ambiguous (historically the word "channel" has meant a lot of things with Meshtastic, people may be unfamiliar with either MQTT or Meshtastic or both, etc)


GUVWAF:
1. It is indeed only necessary for JSON downlink. I'm not sure why uplink didn't work for you before, but it has nothing to do with adding a channel called "mqtt".
2. I'm not sure what you mean? If you create a channel with uplink or downlink enabled you automatically get an MQTT topic with that name.

docs/software/integrations/mqtt/index.mdx
```diff
@@ -103,12 +105,18 @@ If the message received contains valid JSON in the payload, the JSON is deserial
 
 
 #### JSON downlink to instruct a node to send a message
-You can also send a JSON message to the topic `msh/2/json/mqtt/` to instruct a gateway node to send a message to the mesh. Note that the channel you publish it on **must** be called "mqtt". The JSON message should contain the following fields:
+You can also send a JSON message to a topic like `msh/2/json/mqtt/!abcd1234` (for example) to instruct a gateway node to send a message to the mesh.
```
GUVWAF, line 108: The userID is not needed here. Let's keep it simple and keep it like it was.

docs/software/integrations/mqtt/index.mdx
```diff
@@ -103,12 +105,18 @@ If the message received contains valid JSON in the payload, the JSON is deserial
...
+
+First, ensure that your node has a Meshtastic channel configured called "mqtt" (lowercase.) Enable Downlink. The PSK can be random and doesn't matter. This channel allows the node to listen to messages on the `msh/2/json/mqtt/` topic, but messages will be sent out on LoRa over the default 0 channel.
```
GUVWAF, line 110: It's not necessarily channel 0, you can specify it with the "channel" field.

GUVWAF (changes requested):

docs/software/integrations/mqtt/index.mdx
```diff
@@ -6,6 +6,8 @@ sidebar_position: 6
 description: Bridging mesh networks over the internet and integrating Meshtastic protocols with popular technologies such as Home Assistant, Node Red, and Adafruit IO.
 ---
 
+*Definitions:*
```
GUVWAF, line 9: I think this can be removed?

docs/software/integrations/mqtt/index.mdx
GUVWAF: It also doesn't matter if it's lowercase or not.

zyphlar:
@GUVWAF thanks for your comments! 

You said "If you create a channel with uplink or downlink enabled you automatically get an MQTT topic with that name" and that's fine, I understand that a LongFast topic will exist on MQTT when you enable uplink or downlink. This PR is mostly about the requirement to also create a Meshtastic channel called mqtt, since that needs to be fully specified. As you can tell it's easy to believe a lot of things that may be untrue if they're not accurately documented. 

I'll edit according to your feedback, and when we agree it's perfect we can squash the commits. Thanks for your help! I'm also unsure why mqtt was buggy for me.

GUVWAF:
I see, yes, I agree the process to create a channel called "mqtt" can be worded more verbose.

GUVWAF (approved):

PR #1334, opened 2024-07-11
Update local-groups.mdx

fourteevee:
Added Tucson Meshtastic to this list


fourteevee:
> > Added Tucson Meshtastic to this list
> 
> The name will need to be updated to be in compliance with trademark policies. https://meshtastic.org/docs/legal/licensing-and-trademark/#social-media

Right. Mian just mentioned that to me haha. We'll poll a name within the server and update this commit within a little while.

PR #1336, opened 2024-07-13
Reflect new meshtastic versions of Seeed 1110

fifieldt:
Since the original docs were made, Seeed released a specific Metastatic version of the Wio Tracker 1110, the "Wio Tracker 1110 Dev Kit for Meshtastic".

Update the documents to select this tab by the default, and change the store page so people looking to buy get the Meshtastic version rather than the vanilla one.

### DIFF
--- a/docs/hardware/devices/seeed-wm1110/index.mdx
+++ b/docs/hardware/devices/seeed-wm1110/index.mdx
@@ -12,19 +12,20 @@ import TabItem from "@theme/TabItem";
 <Tabs
 groupId="wm1110"
 queryString="wm1110"
-defaultValue="wio-sdk-wm1110"
+defaultValue="wio-tracker-wm1110"
 values={[
 {label: 'WM110 Dev Kit', value:'wio-sdk-wm1110'},
-{label: 'WM110 Tracker', value: 'wio-tracker-wm1110'},
+{label: 'Wio Tracker 1110', value: 'wio-tracker-wm1110'},
 ]}>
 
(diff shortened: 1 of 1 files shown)


docs/hardware/devices/seeed-wm1110/index.mdx
```diff
@@ -67,11 +68,7 @@ Basic support for this device has only just been added, and not all features wor
 
 <TabItem value="wio-tracker-wm1110">
 
-## Seeed Wio Tracker 1110
-
-:::caution Firmware Version Notice
-Basic support for this device has only just been added, and not all features work
-:::
+## Wio Tracker 1110 Dev Kit for Meshtastic
```
tekstrand, line 71: Being picky here, I know, but this doesn't need `for Meshtastic` on the end. We're already on the meshtastic website and since this shows up in the sidebar it's better to keep it shorter. It will also match the title from the other section

tekstrand (approved):
Pre approving, if you can fix up that heading I'm good with merge 👍

docs/hardware/devices/seeed-wm1110/index.mdx
fifieldt: @tekstrand , there is a similarly named but different product, the "Wio Tracker 1110 Dev Board". I used the full name of the product here to distinguish the two.

fifieldt: Feel free to fix with that in mind, otherwise I'll get to it next week.

PR #1338, opened 2024-07-13
Add Seeed images

rcarteraz:
Adds an image for both Seeed Dev Kits


tekstrand (commented):
Why three image files but only two used in the docs?

tekstrand (approved):
👍

PR #1345, opened 2024-07-18
Update installation.mdx

SimmerV:
Added pipx as an alternative to pip3


pdxlocations:
I'm not familiar with pipx.  @ianmcorvidae any insight into this?

ianmcorvidae:
`pipx` is great, and I'd definitely like it to be in the docs for this. Really, it's better in many ways than using normal `pip`, but as a starting point, at least mentioning it as a solution to the `externally-managed-environment` error will let people know about how to fix that (it's common, for example, on raspberry pi installs).

To provide a little context, `pipx` basically works by automatically creating virtual environments for any given thing you install with it, so as to not need to touch system python libraries & keep everything nicely separated. It's pretty much a way of using a venv without having to actually manage a venv yourself (which isn't _too_ hard, but is more of a developer thing, where pipx is more of a user thing).

pdxlocations (approved):

PR #1352, opened 2024-08-01
Update iOS offline map instructions

elliott-liu:
There seems to be a lot of confusion ([1](https://meshtastic.discourse.group/t/offline-maps-on-meshtastic/137/37), [2](https://meshtastic.discourse.group/t/offline-map-data/10504/4), [3](https://www.reddit.com/r/meshtastic/comments/17h2e58/ios_app_how_to_force_the_app_to_download_and_only/)) based on the previous instructions for offline maps.

I've gone ahead and updated the instructions with working instructions of the current implementation, and moved the previous instructions under "Legacy Mesh Map".

I'm not 100% sure it was app version 2.2.10 that added the toggle  but going through the commit history this is the earliest known mention of the term.

### DIFF
--- a/docs/software/apple/usage.mdx
+++ b/docs/software/apple/usage.mdx
@@ -7,7 +7,15 @@ sidebar_position: 2
 
 ## Offline Maps
 
-The Meshtastic app for iOS, iPadOS and macOS supports the sharing of a .mbtiles file with the app for offline map support.
+### Apple Maps
+
+By default, the Meshtastic app for iOS supports the use of native offline Apple Maps.
+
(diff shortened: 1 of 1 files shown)


garthvh (commented):
Mbtiles support is gone, the offline tiles will remain until iOS 18 is released

elliott-liu:
As far as I can tell the default settings of the iOS app is not using .mbtiles. To use .mbtiles you need to toggle on "Legacy Mesh Map" in Settings.

I've moved the original .mbtile instructions under a heading "Legacy Mesh Map" to make things clearer, and added instructions for the default setting (using offline Apple Maps).

I was hoping this would clear up a some confusion, or at least make things a bit more clearer.

elliott-liu:
Once iOS 18 is released .mbtiles the "Legacy Mesh Map" in the documentation can be removed or adjusted as needed.

garthvh (commented):
Not iOS 18 dependent, was removed yesterday

PR #1369, opened 2024-08-13
Fix Blog Description

rcarteraz:
Moves it to the correct spot


tekstrand (changes requested):
Should remove these attributes from the navbar -> items as well instead of just adding here

tekstrand (approved):
looks good!

PR #1379, opened 2024-08-18
docs: update tips on creating default secondary channels

nickpascucci:
The tips on creating a secondary channel which is compatible with the default primary channel said that the default channel name was "LongFast", when in fact it's the empty string.

I discovered this because after I had followed these instructions I wasn't able to talk to default nodes. I scanned one of the example config QR codes [here](https://github.com/meshtastic/meshtastic/tree/master/static/img/configuration/qr-private-primary-example/) and dumped the config with `meshtastic --info`, and saw that the mismatch was due to the channel name.


garthvh:
This is not accurate, you can't use empty string as the name for a secondary channel using the default key

ianmcorvidae:
The completely default name is blank, but in firmware that's changed to a name based on the modem preset, and not all clients support literally entering a blank name (particularly on a secondary channel). `LongFast` is the correct name to have in the documentation here. If you were unable to talk to nodes on default settings, something else is wrong, most likely the frequency slot, or an incorrectly capitalized or otherwise altered name.

PR #1392, opened 2024-08-23
Add gitpod as a compiling option

Nestpebble:
Gitpod is a clean way to compile the firmware for casuals like me, without installing anything locally.


Nestpebble:
Some really good comments - I'll update the instructions if/when https://github.com/meshtastic/firmware/pull/4551 is reviewed/accepted.

PR #1402, opened 2024-08-28
Add warning about neighbor info shared positions

Forge36:


ianmcorvidae:
NeighborInfo packets don't contain or send any positions -- see https://buf.build/meshtastic/protobufs/docs/main:meshtastic#meshtastic.Neighbor and the following entry (NeighborInfo) for what's actually being sent. I'm not sure what you're seeing but it's not this, I don't think.

GUVWAF:
This is incorrect, there is no position data in the NeighborInfo.

PR #1408, opened 2024-08-31
Improve the timezone documentation

ketan:
The current document is confusing if one should enter the TZ name or the "code" corresponding to the zone.


todd-herbert:
Just something else I stumbled across: I'm not sure if all platforms support daylight saving. I have a vague memory of NRF52 ignoring DST changes. I don't think it was a Meshtastic firmware issue though; something deeper about the platform.

PR #1417, opened 2024-09-03
Modify css to reduce spacing

pdxlocations:
This PR modifies the CSS to reduces spacing between lines, after images, and in lists.
[preview link](https://sandbox-git-reduce-text-and-image-spacing-pdxlocs-projects.vercel.app/docs/configuration/tips/)


src/css/custom.css
```diff
@@ -271,4 +273,11 @@ blockquote p::after,
 .markdown blockquote p::after {
   content: none !important;
   display: none !important;
+}
+
+p{
+  line-height: 1.5;
+}
```
tekstrand, line 280: 1.6 should probably be the minimum here
```suggestion
p {
  line-height: 1.6;
}
```

src/css/custom.css
```diff
@@ -271,4 +273,11 @@ blockquote p::after,
...
+.markdown img {
+  margin-bottom: 0; /* Adjust this value as needed */
```
tekstrand, line 282: This will be half of the current value, 0 was a bit too close in my opinion.
```suggestion
  margin-bottom: 1rem;
```

src/css/custom.css
```diff
@@ -14,6 +14,7 @@
   --ifm-list-item-margin: 0;
   --ifm-code-font-size: 95%;
   --ifm-z-index-fixed: 1;
+  --ifm-leading: .8rem;
```
tekstrand, line 17: Just curious how you landed on this number. Wondering if it's worth just keeping it as 1rem so it lines up with the other standard spacing? Either way is fine, they both work. I am more curious than anything else

pdxlocations (commented):
I'm confused by the image margin css

tekstrand (approved):
Didn't mess around with the img margin but in my opinion it's not a big deal.

PR #1418, opened 2024-09-03
Rework erasing and recoverig bootloaders

caveman99:


caveman99:
n.b. the links for the generic bootloaders will only work when https://github.com/meshtastic/firmware/pull/4612 is merged

thebentern (approved):

PR #1419, opened 2024-09-04
Update FrequencyCalculator.tsx

WOD-MN:
adding nepal as a new region


WOD-MN:
add nepal as new region

WOD-MN:
add nepal as new region

caveman99:
You can't add a region to the calculator that is not in the actual firmware, sorry.

WOD-MN:
how to add the region in the firmware? wanna add the Nepal as a new regin with 86-868 mhz and 100% duty cycle with 30dBm ! it's free frequency band for IoT in Nepal

PR #1436, opened 2024-09-08
fix: update the factory erase documentation

nhecker:
This PR updates documentation which was not completely correct. Alternatively, the referenced `uf2` files should be made available in the Stable zip, but that's beyond the scope of this PR which is intended to correct the documentation only.

### DIFF
--- a/docs/getting-started/flashing-firmware/nrf52/nrf52-erase.mdx
+++ b/docs/getting-started/flashing-firmware/nrf52/nrf52-erase.mdx
@@ -20,7 +20,7 @@ Alternatively, follow the instructions below.
 
 To reset the flash storage on your nRF52 board:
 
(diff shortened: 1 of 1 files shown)


caveman99:
This was changed a few days ago to reflect the new versions of the tools. The new tools will be available in all new firmware builds from now on. No need to change the wording here and change it back next week.

nhecker:
As of today, only `Meshtastic_nRF52_factory_erase_v2.uf2` is available in the Stable build's zip. I had to go digging in the Alpha build's zip for the `Meshtastic_nRF52_factory_erase_v3_S140_6.1.0.uf2` and `Meshtastic_nRF52_factory_erase_v3_S140_7.3.0.uf2` which are referenced by the documentation. Sounds like those will be included in the Stable build soon, though.

PR #1454, opened 2024-09-14
Heltec Wireless Tracker Split

rcarteraz:
Heltec Wireless Tracker has been separated into **v1.0** and **v1.1** due to different firmware files, and different selections in the flasher. This way people don't get confused.


tekstrand (approved):
Changes lgtm

PR #1469, opened 2024-09-26
Update faq.mdx

jose1711:
Improve the message icons meaning.

### DIFF
--- a/docs/about/faq.mdx
+++ b/docs/about/faq.mdx
@@ -44,8 +44,8 @@ export const Faq = {
       content: `
   - Cloud with an up arrow - Queued on the app to be sent to your device.
   - Cloud only - Queued on the device to be sent over the mesh.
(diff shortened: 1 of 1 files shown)


GUVWAF:
For a message to a channel, you'll never get the person with checkmark, so "Note that you should only consider the message delivered if it was acknowledged by the recipient" isn't exactly the case.

garthvh:
These seem to add confusion

Nestpebble:
Rather than describe the icons, can we embed them, and use the description as alt-text?

jose1711:
> Rather than describe the icons, can we embed them, and use the description as alt-text?

That would be great. It would also be helpful to indicate that a direct message (DM) without a check mark likely means the message wasn't delivered successfully, assuming all messages are waiting for acknowledgment.

PR #1471, opened 2024-09-28
Update region-by-country.mdx

jose1711:
Fix typos.


tekstrand:
I believe these are just localized spellings, not typos

jose1711:
> I believe these are just localized spellings, not typos

Currently those look like a hybrid between German (Slowakei/Slowenien) and English spelling (Slovakia/Slovenia).

PR #1472, opened 2024-09-28
Update local-groups.mdx

KG5CNK:
Added Louisiana group with contact information to list


pdxlocations (changes requested):
Hello, I'm not sure an email address would be considered a 'local group'.  Do you have a facebook page, discord server, etc that might be more appropriate?

Also we won't support groups that are called "Meshtastic [location]" as they could be misinterpreted as being officially endorsed.  There are guidelines and suggestions on the trademark page:

pdxlocations:
Closing due to no response to change request.

PR #1481, opened 2024-10-01

Xaositek:
Following discussion on Discord regarding S&F out of the box operations, these updates may provide useful assistance in getting the system operational for new users


GUVWAF:
I'm still not sure we should document how to circumvent the limitation of not being able to request it on the public channel. 
The implementation in the firmware is not perfect, that's why it still works with the CLI and if you're lucky (last NodeInfo you received from the S&F server was not on the public channel), but it's intended to only use it in private networks, ideally on a different frequency slot, but also just because there will be less traffic there.

Xaositek:
I debated Requirements or Usage Overview, I like the breakdown you have proposed.

Going one step further, should the client specific items go beneath each tabbed section?

Xaositek:
I made several edits to this - now looking for a re-review of these proposals.

PR #1491, opened 2024-10-07
Update local-groups.mdx

xtrcv:
Added Meshastic Lithuania Signal group


tekstrand (changes requested):
Hi @xtrcv thanks for the PR! You'll need to update both the group name and logo to be in compliance with the trademark guidelines

Something like "Lithuania Mesh Group" would work
and you'll need to use your own logo, or the community M Powered logo, instead of the official meshtastic logo.

Thanks!

tekstrand (approved):
Looks good! Thanks for your contribution!

PR #1492, opened 2024-10-07
docs: clarify which keys config_ok_to_mqtt applies to

logikal:
`config_ok_to_mqtt` [only applies to the `defaultpsk` and `eventpsk`](https://github.com/meshtastic/firmware/blob/1c54388bb897ac2a5617fc9a85e1b045265826e3/src/mqtt/MQTT.cpp#L547-L548).

This might be a little verbose because I state what it does apply to and consequences for other keys, but sometimes being that explicit helps. I'm not tied to it.

### DIFF
--- a/docs/configuration/radio/lora.mdx
+++ b/docs/configuration/radio/lora.mdx
@@ -119,7 +119,7 @@ Setting this to option to 'true' means the device will ignore any messages it re
 
 Acceptable values: `true`, `false`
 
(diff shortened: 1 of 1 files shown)


tekstrand (approved):
Anywhere else this might need to change?

pdxlocations:
All appears to be correct to me, but above my pay grade to approve.

PR #1404, opened 2024-08-30
Restructure Hardware Pages

rcarteraz:
# Restructure Hardware Section and Add New Hardware

This PR implements the changes which will close [Reorganize Hardware Section](#1339). It also includes the addition of new Seeed Studio hardware. The goal is to condense and reorganize our hardware section for improved navigation and easier management.

## Changes

- Restructured the hardware menu to show manufacturers at the top level
- Created individual index pages for each manufacturer
- Moved existing product pages under their respective manufacturer index pages
- Updated internal links to reflect the new structure
- Added new Seeed Studio hardware: SenseCAP Card Tracker T1000-E
- Added Heltec Vision Master series hardware
- Added redirects for changes

## New Structure

```
Hardware
└── Devices
    ├── RAK Wireless
    │   ├── WisBlock
    ├── LILYGO
    │   ├── T-Beam
    │   ├── T-Echo
    │   ├── T-Lora
    │   ├── T-Deck
    │   └── T-Watch
    ├── HELTEC
    │   ├── LoRa 32
    │   ├── Sensors
    │   └── Vision Master
    ├── B&Q Consulting
    │   ├── Nano Series
    │   └── Station Series
    ├── Raspberry Pi
    │   └── Pico
    ├── Linux Native Devices
    ├── Seeed Studios
    │   ├── Wio-WM1100
    │   └── SenseCAP
    │       └── Card Tracker T1000-E
    ├── Canary
    ├── unPhone
    └── Chatter
```

Feedback and suggestions for improvement are welcome!

PR #1366, opened 2024-08-13
Add caution about ESD protection on Heltec LoRa 32 V3.

HAL9000COM:
Heltec LoRa 32 V3 does not have proper ESD protection for its USB port and CP2102. I've broke 2 boards due to static discharge. A caution message would be helpful.
The official schematic shows no protection diode, which is recommend in CP2102 datasheet.
Screenshot 2024-08-13 210356
Screenshot 2024-08-13 210413

PR #1493, opened 2024-10-08
Add caution about Heltec LoRa 32 V3 back.

HAL9000COM:
I just noticed that the caution message added in #1366 was removed in #1404 without clear explanation.
This pull request will add the message back.

### DIFF
--- a/docs/hardware/devices/heltec-automation/lora32/index.mdx
+++ b/docs/hardware/devices/heltec-automation/lora32/index.mdx
@@ -57,6 +57,9 @@ Not recommended because of design issues! Support is being phased out. Use V3 in
 <TabItem value="v3">
 
 ## V3/V3.1
+:::caution
+Be careful when interacting with the USB-C port. This device does not have ESD protection for the CP2102 USB to UART bridge chip.
+:::
 
 :::info
 This device may have issues charging a connected battery if utilizing a USB-C to USB-C cable. It's recommended to use a USB-A to USB-C cable.


HAL9000COM:
> > I just noticed that the caution message added in #1366 was removed in #1404 without clear explanation. This pull request will add the message back.
> 
> The "clear explanation" is this was merely a simple inadvertent removal during a complete overhaul of the entire hardware section which involved modifying/adding 60+ pages of the docs. There isn't a conspiracy here, my apologies for the mistake -- it wasn't the only one I made with the change.
> 
> As for the warning itself, while I generally don't like adding something to the docs merely because of one person's issues, but I suppose this is a decent warning to have. Ideally though we should check where this is an issue with all supported hardware rather than just a single device as this may lead some to believe we have since it's only present on this device.
> 
> Additionally, I am not at all a fan of stacked admonitions. We need to either combine the two into one which is usually preferable, but I'm willing to accept nesting them due to the differing info vs caution. See below:
> 
> ```
> ::::info
> This device may have issues charging a connected battery if utilizing a USB-C to USB-C cable. It's recommended to use a USB-A to USB-C cable.
> 
> :::caution
> Be careful when interacting with the USB-C port. This device does not have ESD protection for the CP2102 USB to UART bridge chip.
> :::
> 
> ::::
> ```

Not a big deal. I believe it's just a small mistake in a huge restructure.

This is the only device that I broke due to ESD, it's rare. Due the large base of supported devices, it would be huge effort to check each datasheet and schematic to find similar design flaws. We may just add these info when someone encountered it.

The nested info block much prettier than the stacked one.

PR #1497, opened 2024-10-10
Add power telemetry settings

noon92:
For some reason these were missing.

### DIFF
--- a/docs/configuration/module/telemetry.mdx
+++ b/docs/configuration/module/telemetry.mdx
@@ -77,6 +77,18 @@ This option is used to configure the interval (in seconds) that should be used t
 
 Default is `1800` seconds (30 minutes).
 
+### Power Metrics Enabled
+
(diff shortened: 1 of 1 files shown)


docs/configuration/module/telemetry.mdx
```diff
@@ -77,6 +77,18 @@ This option is used to configure the interval (in seconds) that should be used t
...
 
+### Power Metrics Enabled
+
+This option is used to enable/disable the sending of power telemetry as gathered by an attached supported voltage/current sensor. Note that this does not need to be enabled to monitor the voltage of the battery.
+
+Default is `false`.
+
+### Power Metrics Interval
+
+This option is used to configure the interval (in seconds) that should be used to send power metrics from an attached supported sensor over the mesh network.
```
pdxlocations, line 88: I like the use of 'try' as written in the protobufs:
>Interval in seconds of how often we should try to send our power metrics to the mesh

Usually we copy directly from the protobuf definitions so the wording stays consistent across platforms.

noon92:
sorry for the delay, missed this

PR #1501, opened 2024-10-12
Fixing indentation in FAQ

exotime:
Previous indentation here caused this documentation to be rendered in a code block. Checked the other FAQ items, look like this was the only one.

image


tekstrand (approved):
Nice. Interesting little bug you found there. Thanks for the fix!

PR #1515, opened 2024-10-21
Add CincyMesh to Kentucky

benyanke:
Half of the greater cincy area is in Kentucky as a border-city, so it properly belongs in both categories. Very minor PR adjusting this.

As an example,  several of the nodes on their website's map are in KY.

https://www.cincymesh.org/guidelines

The river in the middle is the border.

image


pdxlocations:
I believe one link per group is enough.  If you don't want to be locked into a state, you could list under US like:

Screenshot 2024-10-22 at 3 45 25 PM

PR #1518, opened 2024-10-22
Clarify signed messages are for DMs.

Nestpebble:
Also added an extra full-writing of Public Key Cryptography (PKC) to make it clearer and mirror the PSK section


Nestpebble:
@jp-bennett would you mind having a quick look, please?

jp-bennett (approved):

PR #1527, opened 2024-10-25
Update remote-admin.mdx

Dokument:
## What did you change
I added a warning banner for version 2.5+ users that the content of the page will not work. 

## Why did you change it
2.5+ users can still follow the documentation and expect the same result. However, they will not. Bad documentation is worse than no documentation. Warning users about this page being out of date until it is updated. The content is otherwise unchanged to benefit users with older nodes.

### DIFF
--- a/docs/configuration/remote-admin.mdx
+++ b/docs/configuration/remote-admin.mdx
@@ -6,6 +6,10 @@ sidebar_position: 3
 description: An advanced feature which allows remote administration of a device through a secure channel on the Mesh instead of via Bluetooth, Serial, or IPv4.
 ---
 
+:::warning Out of Date
+This page is out of date for firmware 2.5+ and will not work.
+:::
+
 :::caution Disclaimer
 This is an advanced feature that few users should need. Keep in mind that it is possible (if you are not careful) to assign settings to a remote node that cause it to completely drop off of your mesh. We advise network admins have a test node to test settings with before applying changes to a remote node to prevent this.
 :::


pdxlocations:
As far as I've tested, the legacy admin channel still works.  At least it does for my 2.5 nodes admining 2.4 nodes.

If the documentation is wrong, lets work to correct it instead of adding warnings.

pdxlocations:
I've been informed that the legacy admin may indeed not work using a 2.5 node to admin a 2.5 node, not that you would need to.  Apologies!

docs/configuration/remote-admin.mdx
```diff
@@ -6,6 +6,10 @@ sidebar_position: 3
 description: An advanced feature which allows remote administration of a device through a secure channel on the Mesh instead of via Bluetooth, Serial, or IPv4.
 ---
 
+:::warning Out of Date
+This page is out of date for firmware 2.5+ and will not work.
+:::
+
```
pdxlocations, line 12: A warning here seems fair as a short term fix, but perhaps we could word it a little differently?  How about something like:

:::info
Since the introduction of improved PKI-based administration in firmware 2.5, using the legacy admin channel described on this page is no longer supported.
:::

Nestpebble: I'm in the middle of rewriting this page. I'll try and include a suitable signpost when I get to that bit.

PR #1532, opened 2024-10-28
Clarified that managed mode isn't required for admin control

Nestpebble:
Clarified that managed mode isn't required for admin control. Also clarified the wording around admin control/legacy admin channel.


docs/configuration/radio/security.mdx
```diff
@@ -34,7 +34,7 @@ The public key(s) authorized to send administrative messages to this node. Only
 
 Acceptable values: `true` or `false`
 
-Enabling Managed Mode restricts access to all radio configurations via client applications. Radio configurations will only be accessible through the Admin channel. Ensure the Admin channel is functioning properly before enabling this mode to avoid being locked out.
+Enabling Managed Mode restricts access to all radio configurations via client applications. This is not required to be set in order to remotely administer the node. Radio configurations will only be accessible through the Admin control messages or the legacy Admin channel. Ensure the Admin control or legacy Admin channel is functioning properly before enabling this mode to avoid being locked out.
```
pdxlocations, line 37: This reads a little confusing to me, how about something like:

Enabling Managed Mode blocks client applications from accessing all radio configurations. However, this setting is not required for remote node administration. Once enabled, radio configurations can only be accessed through Admin control messages or the legacy Admin channel.

Before enabling Managed Mode, verify that the Admin control or legacy Admin channel is working properly to prevent being locked out.

Nestpebble: We'll go with that. Might take me a couple of hours...

pdxlocations: Sorry, re-reading this, would it make more sense to switch it around?

>Enabling Managed Mode blocks client applications from accessing all radio configurations. Once enabled, radio configurations can only be accessed through Admin control messages or the legacy Admin channel. However, this setting is not required for remote node administration.

Nestpebble: Yep, makes more sense. I've changed Admin to Remote Admin, as this is the terminology I'm using on the new Remote-Admin page.

pdxlocations (dismissed):

docs/configuration/radio/security.mdx
```diff
@@ -34,7 +34,9 @@ The public key(s) authorized to send administrative messages to this node. Only
 
 Acceptable values: `true` or `false`
 
-Enabling Managed Mode restricts access to all radio configurations via client applications. Radio configurations will only be accessible through the Admin channel. Ensure the Admin channel is functioning properly before enabling this mode to avoid being locked out.
+Enabling Managed Mode blocks client applications from accessing all radio configurations. Once enabled, radio configurations can only be accessed through Remote Admin control messages or the legacy Admin channel. However, this setting is not required for remote node administration.
```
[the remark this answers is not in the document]
Nestpebble: all accounted for in latest update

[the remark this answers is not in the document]
garthvh: Was this a source of confusion? Not sure the however sentence is necessary if the pkc stuff is cleaned up, I like @pdxlocations phrasing better.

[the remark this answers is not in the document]
Nestpebble: > Was this a source of confusion? Not sure the however sentence is necessary if the pkc stuff is cleaned up, I like @pdxlocations phrasing better.

It came up a few times when I was reading up on the remote_admin update, and seemed a quick fix. You're going to have to be specific about which part of @pdxlocations suggestion you prefer, and which part of the current wording it should replace.

pdxlocations (approved):

PR #1535, opened 2024-10-28
Update Encryption Technical

rcarteraz:
## What did you change
Updated inconsistent use of introducing acronyms. 

## Why did you change it
@Nestpebble Pointed out we weren't consistent.

The best practice is to spell out the acronym the first time it appears in the body text and follow it with the acronym in parentheses. After this initial introduction, you can use the acronym alone. For headings, I don't think introducing an acronym there is best, but I'm willing to change it if others disagree. I think it should be introduced in the body text, even if that means using the acronym alone in a preceding heading. But open to what others think.

### DIFF
--- a/docs/terms/index.mdx
+++ b/docs/terms/index.mdx
@@ -29,7 +29,7 @@ DFU
 : Device Firmware Update, a state which a device is placed into for it to receive a firmware update
 
 ESP32 | [Drivers](/docs/getting-started/serial-drivers/esp32/) | [Firmware](/docs/getting-started/flashing-firmware/esp32/)
-: A chipset of microcontroller made/designed by Espressif, used by a number of devices. Higher power usage than NRF52, but often cheaper and supports WiFi if desired. 
+: A chipset of microcontroller made/designed by Espressif, used by a number of devices. Higher power usage than NRF52, but often cheaper and supports WiFi if desired.
 
 Firmware | [Guide](https://meshtastic.org/docs/getting-started/flashing-firmware/)
 : The low-level software programmed onto a Meshtastic device, controlling its hardware functions and enabling it to communicate within the mesh network using LoRa technology. Firmware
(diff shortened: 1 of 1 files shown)


docs/development/reference/encryption-technical.mdx
Nestpebble, line 1: See my specific comment below. I think either the full words or the abbreviation of DMs can be taken out of header 2, as well.

Overall it's a good tidy-up of the language here.

docs/development/reference/encryption-technical.mdx
```diff
@@ -6,17 +6,17 @@ sidebar_position: 5
...
 ---
 
-## PSK for Channels, PKC for Direct Messages (DMs) and Admin Messages
+## PSK for Channels, PKC for Direct Messages and Admin Messages
 
-Starting with firmware version 2.5.0, which introduces Public Key Cryptography (PKC) for Direct Messages (DMs) and additionally incorporating the use of Session IDs for Admin Messages, the security architecture of Meshtastic has reached a new level of robustness.
+Starting with firmware version 2.5.0, which introduces Public Key Cryptography (PKC) for Direct Messages (DMs) and also incorporates the use of Session IDs for Admin Messages, the security architecture of Meshtastic has reached a new level of robustness.
 
-Below is a detailed overview of how PSK, PKC, and Session IDs are integrated into your system to secure Channels, Direct Messages, and Admin Messages.
+Below is a detailed overview of how Pre-Shared Key (PSK), PKC, and Session IDs are integrated into your system to secure Channels, DMs, and Admin Messages.
```
Nestpebble, line 13: I think there's a sentence missing here, along the lines of:
"Up to version 2.4.x of the firmware, Meshtastic relied entirely on Pre-Shared Key (PSK) cryptography for all messaging."

This serves 3 purposes - it name-checks PSK, it makes it clear that PSK was there all along, and it gets the full tl:dr in the first paragraph.

Nestpebble (changes requested):

Nestpebble (approved):

Nestpebble (approved):

PR #1544, opened 2024-10-31
Add T1000-E to Community Favorites

pdxlocations:
This PR adds the SenseCAP T1000-E to the list of Community Favorites on the Hardware index page


GUVWAF:
While it's certainly a very nifty device, I'm not sure if we should add it here before the incompatibility issues with older SX127x radios is resolved.

pdxlocations:
Makes sense.  It's exciting to have some working-off-the-shelf products so I jumped the gun.  I haven't been following the SX127x issue fully, and thought it had already been resolved with RadioLib updates.  Closing.

PR #1550, opened 2024-11-01
Remove italic "It"

pdxlocations:
Only the word "It" is italicized.  This is weird.  I also changed the wording a tiny bit for clarity.

### DIFF
--- a/docs/getting-started/index.mdx
+++ b/docs/getting-started/index.mdx
@@ -54,7 +54,7 @@ If your device is not listed above, please review our [supported devices](/docs/
 
 :::danger STOP! Put The Power Cable Down!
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/index.mdx
```diff
@@ -54,7 +54,7 @@ If your device is not listed above, please review our [supported devices](/docs/
 
 :::danger STOP! Put The Power Cable Down!
 
-Never power on the radio without attaching an antenna! _It_ could damage the radio chip.
+Never power on the radio without attaching an antenna! Doing so could damage the radio chip.
```
tekstrand, line 57:
```suggestion
Never power on the radio without attaching an antenna as doing so could damage the radio chip!
```

How about this?

tekstrand (approved):
Either way, change lgtm

PR #1553, opened 2024-11-04
Update local-groups.mdx - Update url

ElranRegev:
Replaced URL from Telegram group to new club website!

## What did you change
New website URL replaced from Telegram URL

## Why did you change it
Since we have a new club website!


tekstrand (approved):
Nice, congrats on the website!

PR #1568, opened 2024-11-13
fix link to android usage instead of python

misterpok:
The link for android usage went to python cli page, this PR changes it to the android usage page.


pdxlocations (approved):
nice catch!

PR #1574, opened 2024-11-14
Update Docusarus to Latest Version

rcarteraz:
## What did you change
This updates the docs to use the latest version of Docusaurus 3.6.1
It also fixes the comments on blog posts so that it'll work with the changes.

## Why did you change it
We were using a really old version, 3.1.1

I've tested just about every page, and most links. So far I have not found anything that is broken with these changes.


pdxlocations (approved):

jfirwin (approved):
Things seem to work so I say let's go for it.

PR #1578, opened 2024-11-16
Homepage (minor) Redesign

rcarteraz:
## What did you change
- Changed getting started into a 1, 2, 3 step process
- Reused cards for adding references to the clients w/ links in a new "Get Connected" section.

## Why did you change it
The verbiage in the cards recently changed which caused the spacing to get crazy. Initially I wanted to just fix that but in reality I was never a fan of how we had those to begin with. 

### Before
Screenshot 2024-11-15 at 7 35 10 PM

### After
Screenshot 2024-11-15 at 7 55 30 PM
Screenshot 2024-11-15 at 7 55 44 PM

image


caveman99 (approved):

src/pages/index.tsx
```diff
@@ -161,61 +168,150 @@ function Home() {
...
+            <div className="flex flex-col items-center">
+              <div className="flex items-center gap-2">
+                <CheckCircleIcon
+                  className="h-8 w-8"
+                  style={{ color: "var(--ifm-color-primary)" }}
+                />
+                <h4 className="text-lg font-bold">Step 1</h4>
               </div>
+              <Link to="/docs/hardware/devices/">
+                <h5 className="font-semibold mt-1">Choose a Device &rarr;</h5>
```
tekstrand, line 183: Nitpick: anchor tags (link) cannot have a block level (heading) element inside of them.

tekstrand:
This looks much better and I think is easier to understand. One recommendation though is that the links in Get Started don't really look like links, I think we could make it more obvious they can be clicked

pdxlocations (approved):

tekstrand (approved):
> > This looks much better and I think is easier to understand. One recommendation though is that the links in Get Started don't really look like links, I think we could make it more obvious they can be clicked
> 
> Just pushed an update. How's that?

Fantastic, nice work!

PR #1592, opened 2024-11-20

battlehax:
Instructions for Avahi to make Native devices autodetect on Android when using TCP.


jp-bennett (approved):

vidplace7:
This config is working great here ❤️

pdxlocations:
@battlehax The single quotes in `standalone='no'` do not persist when running this command. They just disappear and the service throws an error.  Changing them to double quotes fixed the issue.  I'm on a mac.

battlehax:
> @battlehax The single quotes in `standalone='no'` do not persist when running this command. They just disappear and the service throws an error. Changing them to double quotes fixed the issue. I'm on a mac.

A couple of people on discord had similar issues with copy/pasting and quotes becoming unicode. I'll update this, and also update the instructions to use an editor instead of tee, to align better with other instructions on the same page. Thanks!

pdxlocations:
FWIW, tee worked great for me once I swapped out the quotes.  It's an elegant way to create the file, assuming it works!

PR #1608, opened 2024-11-30
Update remote-admin.mdx

ndguardian:
Hello! 

I am proposing the modification to the documentation below. If a user has retrieved their public key to set as the admin key for their node via the web client or one of the mobile apps, it will not include `base64:` at the start of the public key, which appears to be a requirement to actually set it. 

In my experience, not including this leads to an error like is seen below, and some users may encounter problems if they simply try to use their public key as obtained via the apps as mentioned. 

Update CLI command notes to mention to append base64: to the front of the admin key if retrieved from something other than the CLI.

## What did you change
I am proposing the modification to the documentation CLI instructions for setting the admin key. If a user has retrieved their public key to set as the admin key for their node via the web client or one of the mobile apps, it will not include `base64:` at the start of the public key, which appears to be a requirement to actually set it. 

This will allow users to know that they need to typecast their public key to base64, which appears to be a requirement for the CLI. If following the instructions and ONLY using the CLI, the `base64:` will be prepended by default, but will (hopefully) prompt users to do so if not getting their key via the CLI.
## Why did you change it
This is based on my own testing and exploration of enabling remote administration on a node, where failure to prepend the snippet led to the following error below.

```
PS C:\Users\me> python -m meshtastic --set security.admin_key MYKEYWITHOUTPREPENDEDBASE64
Connected to radio
Adding 'MYKEYWITHOUTPREPENDEDBASE64' to the admin_key list
Aborting due to: expected bytes, str found
```

I'll try to update the screenshots below if I can get github to reflect my changes in my version of the repo. For some reason, it shows in this PR but not yet in my repo. Hmm...
### Before
image

### After
image

### DIFF
--- a/docs/configuration/remote-admin.mdx
+++ b/docs/configuration/remote-admin.mdx
@@ -115,7 +115,6 @@ Legacy admin is enabled using the Legacy Admin channel option in [Security Confi
    meshtastic --set security.admin_key "PASTEPUBLICKEYHERE"
    ```
 6. You may add up to 3 Admin Keys, enabling control from up to 3 different nodes.
-7. Note: If you retrieve your Public Key via a mobile app or the web client, add `base64:` to the front of your key. It will look like this. `base64:PASTEPUBLICKEYHERE`
 
 #### Setting up Remote Admin Using the Legacy Method
 
--- a/docs/configuration/remote-admin.mdx
+++ b/docs/configuration/remote-admin.mdx
@@ -115,6 +115,7 @@ Legacy admin is enabled using the Legacy Admin channel option in [Security Confi
    meshtastic --set security.admin_key "PASTEPUBLICKEYHERE"
    ```
 6. You may add up to 3 Admin Keys, enabling control from up to 3 different nodes.
+7. Note: If you retrieve your Public Key via a mobile app or the web client, add `base64:` to the front of your key. It will look like this. `base64:PASTEPUBLICKEYHERE`
 
 #### Setting up Remote Admin Using the Legacy Method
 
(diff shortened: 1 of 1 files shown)


fifieldt (approved):

docs/configuration/remote-admin.mdx
```diff
@@ -115,6 +115,7 @@ Legacy admin is enabled using the Legacy Admin channel option in [Security Confi
    meshtastic --set security.admin_key "PASTEPUBLICKEYHERE"
    ```
 6. You may add up to 3 Admin Keys, enabling control from up to 3 different nodes.
+7. Note: If you retrieve your Public Key via a mobile app or the web client, add `base64:` to the front of your key. It will look like this. `base64:PASTEPUBLICKEYHERE`
```
pdxlocations, line 118:
```suggestion
```

docs/configuration/remote-admin.mdx
```diff
@@ -115,6 +115,7 @@ Legacy admin is enabled using the Legacy Admin channel option in [Security Confi
    meshtastic --set security.admin_key "PASTEPUBLICKEYHERE"
```
pdxlocations, line 115:
```suggestion
   meshtastic --set security.admin_key "base64:PASTEPUBLICKEYHERE"
```

pdxlocations: would this change be enough to clarify that 'base64' needs to be present?

ndguardian: Possibly, and I'll leave it at your discretion. I considered that originally, but if you do retrieve your key via the CLI, it will already have the `base64:` prepended onto the key, and obviously you don't want to do something like `base64:base64:PASTEPUBLICKEYHERE` as a result.

Hopefully folks would be able to intuit that and so it would be a moot point, but...frankly I don't know. Thoughts?

pdxlocations: I'd like to think the average user is smart enough to solve that one.  Personally I find less words = more better.

ndguardian: And applied!

pdxlocations (approved):

PR #1611, opened 2024-12-01
Update Apple OTA Firmware flashing instructions with recommended settings for the nRF DFU app to prevent update failures

asatnik11:
Updated the Apple OTA Firmware flashing documentation to include a caution notice with instructions to change settings in the nRF Device Firmware Update app to prevent failure.

## What did you change
Modified the Apple section of the nRF52/RP2040 over-the-air Firmware update documentation to include a caution. This caution includes suggested settings to prevent firmware update failure when using the nRF Device Firmware Update app. It also mentions that re-flashing the firmware via USB will be required in case of a failed OTA flash.

## Why did you change it
I changed it because going on the roof is no fun.

### DIFF
--- a/docs/getting-started/flashing-firmware/nrf52/ota.mdx
+++ b/docs/getting-started/flashing-firmware/nrf52/ota.mdx
@@ -45,6 +45,16 @@ OTA firmware updates are available for Android using an older release of the mor
 
 OTA firmware updates are available on iOS & iPadOS using the nRF Device Firmware Update App available through the [Apple App Store](https://apps.apple.com/us/app/nrf-device-firmware-update/id1624454660)
 
+:::caution
+
+To ensure successful OTA DFU firmware flashing on iOS, change the following settings in the nRF DFU App:
+- Packet receipt notification: On
+- Number of Packets: 10
+
+Failure to change these settings may cause the OTA firmware update to fail. Failure will result in the the Meshtastic firmware not booting and will require the firmware to be re-flashed via USB.
+
+:::
+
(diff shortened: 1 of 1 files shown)


docs/getting-started/flashing-firmware/nrf52/ota.mdx
```diff
@@ -45,6 +45,16 @@ OTA firmware updates are available for Android using an older release of the mor
 
 OTA firmware updates are available on iOS & iPadOS using the nRF Device Firmware Update App available through the [Apple App Store](https://apps.apple.com/us/app/nrf-device-firmware-update/id1624454660)
 
+:::caution
```
tekstrand, line 48: Wondering if instead of putting this in a callout(caution) we should just include this in the step by step instructions

asatnik11: If you'd prefer to include the suggested nRF DFU app settings in the step-by-step instructions it still gets the point across. I still think it makes sense to include a caution at the top of the page, above the Android/Apple tabs, regarding the need to re-flash firmware via USB if the OTA flash fails. I think it could prevent a lot of frustration.

I'm not very experienced with GitHub pull requests - Are you able to briefly guide me with the best way to amend my pull request with additional changes to the same file? If you don't have time or don't want to, I completely understand.

garthvh: > Wondering if instead of putting this in a callout(caution) we should just include this in the step by step instructions

Agree with this no need to scare every user

PR #1612, opened 2024-12-02
update OTA instructions for iOS with packet settings

hayschan:
## What did you change

- Updated the **iOS OTA firmware update** section of the documentation to include additional troubleshooting steps for resolving upload failures.
- Added instructions for enabling "Packets Receipt Notification" in the nRF DFU app, allowing users to adjust the "Number of Packets" setting. This resolves common issues with slow or failed Bluetooth OTA updates.
- Clarified the expected upload time (~1 kB/s, approximately 10 minutes) and troubleshooting steps for improving upload stability.
- Added a reminder to the iOS OTA firmware update section about preventing the screen from turning off during the firmware upload.
- Recommended temporarily setting the iPhone's auto-lock to "Never" to ensure the phone stays awake and the upload process isn't interrupted by screen timeout.

## Why did you change it

- Multiple users, including myself, experienced difficulties with Bluetooth OTA updates. After adjusting the packet number in the nRF DFU app, the process successfully completed.
- This issue has been reported on community platforms like Discourse and Reddit, and adjusting the packet number was a widely shared solution.
- The iPhone's default auto-lock settings (30 seconds or 1 minute) can cause the phone's screen to turn off during Bluetooth OTA uploads, potentially interrupting the process.

### Before

_No specific issues with the previous documentation._

### After

- Added troubleshooting steps for enabling "Packets Receipt Notification" and adjusting packet numbers.


docs/getting-started/flashing-firmware/nrf52/ota.mdx
```diff
@@ -51,5 +51,17 @@ OTA firmware updates are available on iOS & iPadOS using the nRF Device Firmware
...
+    1. Enable "**Packets Receipt Notification**" in the app settings. 
+    2. After enabling this, the "**Number of Packets**" setting will become active, allowing you to reduce it. Try changing the number of packets to **5**. 
+    3. Upload the firmware. Wait patiently. The upload speed is reported to be around 1 kB/s, so the process may take approximately 10 minutes to complete.
+    4. If it still fails, reduce the value by one and retry the upload.
+3. **Prevent screen timeout during upload**: The iPhone's auto-lock feature is often set to 30 seconds or 1 minute by default. When this happens, the screen will go black, and the phone may go to sleep, potentially interrupting the Bluetooth firmware upload. To avoid this, temporarily set the auto-lock to "Never" during the upload process to ensure that the phone stays awake and the upload completes without interruption.
+
+*Note: Enabling "Packets Receipt Notification" is critical, as it unlocks the ability to modify the packet count. Reducing the packet size can significantly improve the upload stability and reduce the likelihood of failure.*
```
pdxlocations, line 64:
```suggestion
```
This seems unnecessary since it's already written above

hayschan: Good point. I will remove this as suggested.

pdxlocations:
@rcarteraz @tekstrand could one of you verify this one?  I've had good luck just sticking with defaults when using the DFU app, with speeds around 4kB/s.

hayschan:
> @rcarteraz @tekstrand could one of you verify this one? I've had good luck just sticking with defaults when using the DFU app, with speeds around 4kB/s.

Some context of the device I am using.

- the Meshtastic hardware: Heltec T114 v2 433-510 Mhz version. Equipped with nRF52
- Phone: iPhone 13 Pro and iPhone SE (Gen 2)

The default number of packets, 23, does not work on both of my iPhone. 10 also does not work, stuck forever and eventually lead to a timeout. Eventually, it was 5 that worked for me. I did not try the number between 5 to 10, though.

tekstrand:
> @rcarteraz @tekstrand could one of you verify this one? I've had good luck just sticking with defaults when using the DFU app, with speeds around 4kB/s.

I have had success with the default settings, although I can't say I have done extensive testing.

If I am understanding correctly, we got these recommended values through trial and error - if that is true, I don't feel confident about adding them to our docs. I think we need to validate with near 100% certainty that they are correct.

hayschan:
> > @rcarteraz @tekstrand could one of you verify this one? I've had good luck just sticking with defaults when using the DFU app, with speeds around 4kB/s.
> 
> I have had success with the default settings, although I can't say I have done extensive testing.
> 
> If I am understanding correctly, we got these recommended values through trial and error - if that is true, I don't feel confident about adding them to our docs. I think we need to validate with near 100% certainty that they are correct.

100% agree with your point. That's exactly why the change was not made as part of the regular steps. Instead, the recommendation is included as a separate tip after the regular steps. This ensures that it's presented as an optional troubleshooting step, specifically for users who may experience issues with the default settings.

> ... the regular step content
> 
> If the upload fails, here are some tips based on common issues:
> 1. **Ensure the correct app is used:** Double-check that you’ve downloaded the **nRF Device Firmware Update (nrF DFU)** app, not the nRF Connect app.  
> 2. **Adjust packet settings:** In the app’s settings, the default number of packets is often set to 23. If you experience upload failures, follow these steps to adjust the packet size:

This way, the packet adjustment is offered as a *possible solution* without presenting it as the main method. It’s a tip for those who may encounter issues, rather than a prescribed solution for everyone.

pdxlocations:
> This way, the packet adjustment is offered as a _possible solution_ without presenting it as the main method. It’s a tip for those who may encounter issues, rather than a prescribed solution for everyone.

What we're trying to avoid is catering to the 0.1% of users with highly-specific issues.  That said, I think it would be a nice addition to mention there are settings one could adjust if the upload fails.  As tekstrand mentioned, specifying any recommended values would have to be thoroughly vetted through testing and/or nRF documentation.

b8b8:
Screenshots of my successful settings for iOS and Android.
signal-2024-12-20-084911
signal-2024-12-20-085037

docs/getting-started/flashing-firmware/nrf52/ota.mdx
```diff
@@ -51,5 +51,16 @@ OTA firmware updates are available on iOS & iPadOS using the nRF Device Firmware
...
+2. **Adjust packet settings:** In the app’s settings, the default number of packets is often set to 23. If you experience upload failures, follow these steps to adjust the packet size:
+    1. Enable "**Packets Receipt Notification**" in the app settings. 
+    2. After enabling this, the "**Number of Packets**" setting will become active, allowing you to reduce it. Try changing the number of packets to **5**. 
+    3. Upload the firmware. Wait patiently. The upload speed is reported to be around 1 kB/s, so the process may take approximately 10 minutes to complete.
+    4. If it still fails, reduce the value by one and retry the upload.
+3. **Prevent screen timeout during upload**: The iPhone's auto-lock feature is often set to 30 seconds or 1 minute by default. When this happens, the screen will go black, and the phone may go to sleep, potentially interrupting the Bluetooth firmware upload. To avoid this, temporarily set the auto-lock to "Never" during the upload process to ensure that the phone stays awake and the upload completes without interruption.
+
+
```
pdxlocations, line 64:
```suggestion
The iPhone's auto-lock feature could potentially interrupt the Bluetooth firmware upload. To avoid this, occasionally tap on your screen, or temporarily set the auto-lock to "Never" during the upload process to ensure that the phone stays awake and the upload completes without interruption.

If the update fails, you may find that adjusting the packet settings may help:

1. In settings, enable "**Packets Receipt Notification**". 
2. Change "**Number of Packets**" to a lower value. Some users report success with "5". 
```

pdxlocations: Here's my stab at making these useful instructions a bit more concise, while not making any promises.

tekstrand (approved):

PR #1613, opened 2024-12-03
Suggest against SX1276 devices

turnrye:
## What did you change

I updated the docs to make the wording firmer, clarifying that at this point Semtech SX1276 devices are a generation out of date.

## Why did you change it

I noticed that maintainers were actively advising people not to buy devices using SX1276, which prompted me to realize that the wording wasn't as direct as it should be on the docs.

### DIFF
--- a/docs/hardware/devices/index.mdx
+++ b/docs/hardware/devices/index.mdx
@@ -24,7 +24,7 @@ Please do your research and choose the board that meets your needs (or maybe alr
 
 :::info
 
-- The Semtech SX1262 transceiver is newer than the SX1276 and provides increased receive and transmit performance.
+- Devices using the previous-generation Semtech SX1276 are no longer recommended due to poor performance .
 - nRF52-based devices use a fraction of the power compared to ESP32-based devices and are therefore generally preferred for solar and handset applications.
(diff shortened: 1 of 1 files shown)


docs/hardware/devices/index.mdx
```diff
@@ -24,7 +24,7 @@ Please do your research and choose the board that meets your needs (or maybe alr
 
 :::info
 
-- The Semtech SX1262 transceiver is newer than the SX1276 and provides increased receive and transmit performance.
+- Devices using the previous-generation Semtech SX1276 are no longer recommended due to poor performance .
```
pdxlocations, line 27:
```suggestion
- Devices using the previous-generation Semtech SX1276 are no longer recommended due to their inferior performance.
```

pdxlocations (approved):

GUVWAF:
I'm sorry but I wouldn't call it poor performance, that's just not true. You can still transmit and receive over 100km with it, according to the data sheet for some LoRa settings it has even better receive sensitivity.
It also doesn't suffer from false preamble detections like the SX126x.

In my opinion the old wording was fine.

PR #1616, opened 2024-12-05
A Blog Post for Using Emojis 

rcarteraz:
## What did you change
New blog post on using emojis in the short name, as well as mentioning waypoints. 

## Why did you change it
They're fun!


pdxlocations (approved):
Looks great and answers the common question: "How do I use an emoji for my icon?"

One nit-pick: the image names could be more descriptive since they're all in the `blog` folder.  IE `emoji-01.webp` or `2024_12-image-01.webp`.  This folder will eventually be pretty busy so some unique identifiers to match them with a specific blog post coult be usefull.

thebentern (approved):

PR #1617, opened 2024-12-06
typo: correct `py -V`

Love4yzp:
## What did you change
correct `py -V` to `python -V` for Windows

## Why did you change it
`py -V` is meaningless

### DIFF
--- a/docs/software/python-cli/installation.mdx
+++ b/docs/software/python-cli/installation.mdx
@@ -163,7 +163,7 @@ values={[
 - Check that your computer has Python 3 installed.
   - Use the command
     ```powershell
-    py -V
+    python -V
     ```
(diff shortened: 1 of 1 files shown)


docs/software/python-cli/installation.mdx
```diff
@@ -163,7 +163,7 @@ values={[
 - Check that your computer has Python 3 installed.
   - Use the command
     ```powershell
-    py -V
+    python -V
```
tekstrand, line 166: My understanding is that `py` is available by default when installing on Windows. If it is not working there is either a problem with the installation/path or the install was customized in some way.

Love4yzp: yes, I use virtual env for python.

PR #1618, opened 2024-12-07
fix(docs): Correct rotary encoder event commands in Canned Message Module Documentation

hayschan:
This pull request fixes an issue in the Canned Message Module documentation where incorrect example commands were provided for configuring rotary encoder events. Specifically, the values `KEY_UP`, `KEY_DOWN`, and `KEY_SELECT` were mistakenly used in the example commands instead of the correct enum values: `UP`, `DOWN`, and `SELECT`.  

### Changes Made:
- Replaced all instances of `KEY_UP`, `KEY_DOWN`, and `KEY_SELECT` with the correct enums `UP`, `DOWN`, and `SELECT` in the example command section.

### Why This Fix is Needed:

Using incorrect values such as `KEY_UP` results in errors like:

```shell
canned_message.inputbroker_event_cw does not have an enum called KEY_UP, so you can not set it.
Choices in sorted order are:
    BACK
    CANCEL
    DOWN
    LEFT
    NONE
    RIGHT
    SELECT
    UP
LocalConfig and LocalModuleConfig do not have attribute canned_message.inputbroker_event_cw.
```

### DIFF
--- a/docs/configuration/module/canned-message.mdx
+++ b/docs/configuration/module/canned-message.mdx
@@ -316,9 +316,9 @@ Replace each `GPIO` (x3) below with the GPIO numbers from your hardware setup.
 meshtastic --set canned_message.inputbroker_pin_a GPIO
 meshtastic --set canned_message.inputbroker_pin_b GPIO
 meshtastic --set canned_message.inputbroker_pin_press GPIO
-meshtastic --set canned_message.inputbroker_event_cw KEY_UP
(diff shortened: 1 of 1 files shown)


pdxlocations (changes requested):
Good catch, would you also correct the examples further up the page around line 195?

pdxlocations (approved):

PR #1623, opened 2024-12-13

SpudGunMan:
## What did you change
add note for elevated SSL access

## Why did you change it
to help others remember


pdxlocations (commented):
I'm not understanding what this note means.  What action is needed for the user?

SpudGunMan:
just to have a note, I dont know its intuitive in the example, that because port 443 is in use sudo is required in a lot of cases

pdxlocations:
Sorry, I'm still confused.  What is the command that requires sudo?

SpudGunMan:
this just calls out sudo it sounds like its not handy closing

PR #1628, opened 2024-12-16
Update index.mdx to reflect removal of support

iwtiwt:
Include info that 2.5.15.79da236 Beta removed support for these units in https://github.com/meshtastic/firmware/pull/5436 .

## What did you change
Add information at top of page noting removal of support.

## Why did you change it
https://github.com/meshtastic/firmware/pull/5436 included in release https://github.com/meshtastic/firmware/releases/tag/v2.5.15.79da236 has removed the wireless capsule support, but appears documentation has not been updated to warn people off purchasing unsupported units.

### DIFF
--- a/docs/hardware/devices/heltec-automation/sensor/index.mdx
+++ b/docs/hardware/devices/heltec-automation/sensor/index.mdx
@@ -19,6 +19,10 @@ values={[
 
 ## Heltec Capsule Sensor Rev. 3.0
 
+:::info
+https://github.com/meshtastic/firmware/releases/tag/v2.5.15.79da236 removed support for the Heltec Wireless Capsule Sensor V3.
+:::
+
 - **MCU:**
   - ESP32-S3FN8 (WiFi & Bluetooth)
 - **LoRa Transceiver:**


docs/hardware/devices/heltec-automation/sensor/index.mdx
```diff
@@ -19,6 +19,10 @@ values={[
 
 ## Heltec Capsule Sensor Rev. 3.0
 
+:::info
```
tekstrand, line 22: @rcarteraz @pdxlocations what do y'all think about just removing this from the docs altogether? I don't think we need to keep around documentation for something the project no longer supports.

If we do decide to keep it, let's find a way to make this clear without using an admonition (`:::info`)

pdxlocations: I'm in favor of removing it. Let's not list unsupported hardware.

tekstrand (changes requested):
@iwtiwt please remove the documentation for this altogether rather than adding a note. Thank you!

iwtiwt:
OK, cool, thankyou.

PR #1629, opened 2024-12-17
Remove rak promotion

garthvh:
Remove promotional sales links for RAK products


tekstrand:
Why are we removing these?

garthvh:
No longer necessary

PR #1646, opened 2024-12-30
Add ROUTER_LATE

rcarteraz:
## What did you change
Adds ROUTER_LATE role to list of roles and comparison table. 

## Why did you change it
New role that needs to be documented.

### DIFF
--- a/docs/configuration/radio/device.mdx
+++ b/docs/configuration/radio/device.mdx
@@ -49,7 +49,7 @@ This table shows the **default** values after selecting a preset. As always, ind
 | TAK            | Yes             | Optional       | Regular           | Yes            | No                  | Yes                   |
 | TAK_TRACKER    | Yes             | Optional       | Regular           | Yes            | No                  | Yes                   |
 | ROUTER         | No[^2]          | No             | High              | Yes            | Yes                 | Yes                   |
-| ROUTER_LATE    | Yes             | Yes            | Regular           | Yes            | No                  | Yes                   |
+| ROUTER_LATE    | Yes             | Yes            | High           | Yes            | No                  | Yes                   |
 | REPEATER       | Yes             | No             | High              | Yes            | Yes                 | No                    |
 
 [^1]: **Awake Only** indicates retransmissions will occur normally but not while asleep between transmission intervals.


docs/configuration/radio/device.mdx
```diff
@@ -47,6 +49,7 @@ This table shows the **default** values after selecting a preset. As always, ind
 | TAK            | Yes             | Optional       | Regular           | Yes            | No                  | Yes                   |
 | TAK_TRACKER    | Yes             | Optional       | Regular           | Yes            | No                  | Yes                   |
 | ROUTER         | No[^2]          | No             | High              | Yes            | Yes                 | Yes                   |
+| ROUTER_LATE    | Yes             | Yes            | Regular           | Yes            | No                  | Yes                   |
```
GUVWAF, line 52: I think we should set power consumption to "High" as it will rebroadcast as much as `ROUTER`/`REPEATER`, and it also has BLE/WiFi/Serial still enabled.

garthvh: Why have we left BLE and WiFi on for a router role?

GUVWAF: I think for a rooftop node that's still a valid use case.

GUVWAF (approved):

docs/configuration/radio/device.mdx
```diff
@@ -14,24 +14,26 @@ The device config options are: Role, Serial Output, and Debug Log. Device config
...
+| REPEATER       | Infrastructure node for extending network coverage by always rebroadcasting packets once with minimal overhead. Not visible in Nodes list.                       | Best positioned in strategic locations to maximize the network's overall coverage. Device is not shown in topology.                                         |
+| ROUTER         | Infrastructure node for extending network coverage by always rebroadcasting packets once. Visible in Nodes list.                                                 | Best positioned in strategic locations to maximize the network's overall coverage. Device is shown in topology.                                             |
+| ROUTER_LATE    | Infrastructure node that always rebroadcasts packets once but only after all other modes, ensuring additional coverage for local clusters. Visble in Nodes list. | Ideal for covering dead spots or ensuring reliability for a cluster of nodes where placement doesn’t benefit the broader mesh. Device is shown in topology. |
```
tekstrand, line 29: I'm being picky here but I think we can improve this some. The description column doesn't need to include the purpose, that can be reserved for the Best Uses column.

> Infrastructure node that always rebroadcasts packets ~once but only~ after all other modes ~,ensuring additional coverage for local clusters~. Visble in Nodes list.

tekstrand: No question as to it's accuracy. I still think that in terms of readability for documentation, which is where my contributions on this project have focused, it can match the other descriptions more closely for brevity and clarity.

I won't push any further as it isn't critical, but I did want to make clear the benefits I see.

PR #1648, opened 2024-12-31
Add `mtnme.sh` to local groups.

vidplace7:
Adds [MtnMe.sh](https://mtnme.sh) group.

Attributed to the "United States" section because we cover North GA and East TN.

### DIFF
--- a/docs/community/local-groups.mdx
+++ b/docs/community/local-groups.mdx
@@ -192,6 +192,7 @@ us on [Discord](https://discord.com/invite/ktMAKGBnBs) to add your group.
 ## United States
 
 - [Midwest Mesh](https://discord.gg/wYwD56K439)
+- [Mountain Mesh (North GA / East TN)](https://mtnme.sh)
 
 ### Arizona
 


docs/community/local-groups.mdx
```diff
@@ -192,6 +192,7 @@ us on [Discord](https://discord.com/invite/ktMAKGBnBs) to add your group.
 ## United States
 
 - [Midwest Mesh](https://discord.gg/wYwD56K439)
+- [Mountain Mesh (North GA / East TN)](https://mtnme.sh)
```
tekstrand, line 195: @pdxlocations @rcarteraz what do y'all think about this situation, would it be better for us to just double list it, once under each state?

pdxlocations: It's my opinion that groups only be listed once on the page, so listing under United States would be appropriate in this case.

pdxlocations:
Screenshot 2025-01-01 at 9 28 18 AM

my only concern is that this logo opens a doorway to trademark abuse

vidplace7:
> Screenshot 2025-01-01 at 9 28 18 AM
> my only concern is that this logo opens a doorway to trademark abuse

image
Fixed :+1:

pdxlocations (approved):

PR #1650, opened 2025-01-02
Remove Incorrect Apple PKC Admin Step?

pdxlocations:
This PR removes the instruction #8 on https://meshtastic.org/docs/configuration/remote-admin/#apple:
`On the remote node, got to Settings > App Settings and enable Administration.`

I believe this setting is client-side only and this step doesn't do anything.

image


garthvh (approved):
Yes this step is wrong

PR #1654, opened 2025-01-04
Update convert-rak4631r.mdx to add extra RUI3 steps

eresonance:
Had to do these steps from the official RAK documentation in order to run the steps already present in the guide.

## What did you change
Added some extra steps from here:
https://docs.rakwireless.com/product-categories/wisblock/rak4631-r/dfu/#converting-rak4631-r-to-rak4631
That are needed prior to the steps already in the docs.

## Why did you change it
Took me forever to figure it out how to properly get the bootloader & meshtastic flashed :) I got this RAK4631 from digikey so maybe there are some manufacturing differences from whomever did the docs the first time.

## Screenshots
Rich diff below shows the changes well.

### DIFF
--- a/docs/getting-started/flashing-firmware/nrf52/convert-rak4631r.mdx
+++ b/docs/getting-started/flashing-firmware/nrf52/convert-rak4631r.mdx
@@ -25,8 +25,38 @@ Here are two ways to flash the bootloader:
    ```shell
    adafruit-nrfutil --verbose dfu serial --package ./WisCore_RAK4631_Board_Bootloader.zip  -p /dev/ttyACM0 -b 115200 --singlebank --touch 1200
    ```
-   Note: The serial port name (`/dev/ttyACM0`) may differ depending on your operating system. Make sure to identify the correct port name for your setup.
+   Note: The serial port name (`/dev/ttyACM0`) may differ depending on your operating system. Make sure to identify the correct port name for your setup. For instance on Windows this will be `COM#` with `#` being a random number, check your device manager.
 6. Continue with the normal [flashing instructions](/docs/getting-started/flashing-firmware/nrf52/drag-n-drop)
+
+### Extra RUI3 Steps
+If the above steps fail with errors like:
+
+```
+Touched serial port COM11
+Opened serial port COM11
+Starting DFU upgrade of type 2, SoftDevice size: 0, bootloader size: 39000, application size: 0
+Sending DFU start packet
+Timed out waiting for acknowledgement from device.
+
+Failed to upgrade target. Error is: No data received on serial port. Not able to proceed.
+```
+
+You will need to follow the first part of the [Converting RAK4631-R to RAK4631](https://docs.rakwireless.com/product-categories/wisblock/rak4631-r/dfu/#converting-rak4631-r-to-rak4631) instructions.
+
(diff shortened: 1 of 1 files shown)


pdxlocations:
@eresonance thanks for your contribution to the docs!  I've never had to go through this process so I personally won't be able to review the procedures you've added.  I do however have issue with linking in images that we don't own or control.  I didn't read through the entire PR to sort out the differences, but would it make sense to just add a link to the RAK docs here?  

It sounds like adding or linking this information would save users like you a great deal of time, so thanks again!

tekstrand:
I've gone through this process a number of times and agree with @pdxlocations the best option is to link to the RAK docs. We've made a choice on the Meshtastic docs that we don't duplicate or copy over instructions that already exist elsewhere, so I would agree that still remains the best option here.

eresonance:
I did link to the RAK instructions, but you only need to do up to step 6 and not the rest of them:
https://docs.rakwireless.com/product-categories/wisblock/rak4631-r/dfu/#converting-rak4631-r-to-rak4631
I took the stack overflow approach of having the steps listed in these docs instead of just relying on an outside link, just in case RAK changes their docs. That's somewhat likely as their end of the docs has nothing to do with meshtastic and they're a totally separate entity from you guys.

My recommendation is to keep the steps as is, but remove the images. I'll resubmit the MR.

eresonance:
Updated, new changes look like:

image

docs/getting-started/flashing-firmware/nrf52/convert-rak4631r.mdx
```diff
@@ -25,9 +25,36 @@ Here are two ways to flash the bootloader:
...
+2. Connect the RAK4631-R via USB and check if port has been detected via device manager. The COM port number is not fixed and can be different depending on the PC. If no port is shown in device manager, you can try to double click reset button on the WisBlock Base and check again.
+3. After that, you need to send `AT+BOOT` command to the device via serial. You can follow the guide on using [Tera Term](https://docs.rakwireless.com/product-categories/wisblock/rak4631-r/dfu/#how-to-check-firmware-version-using-tera-term) (or use putty, minicom, etc.) but instead of checking the firmware version, you have to input `AT+BOOT` and hit enter. You will see no reply since the module will restart then will be disconnected momentarily before re-establishing again the connection to Tera Term.
+
+> [!WARNING]
+> You have to disconnect the device connection to Tera Term or close it so that the COM port will be free when you do the firmware update on the next step. Else, you will have error during FW update.
```
pdxlocations, line 51:
```suggestion
> You have to disconnect the device connection to Tera Term or close it so that the COM port will be free when you do the firmware update on the next step, or you will have an error during FW update.
```

pdxlocations:
This seems reasonable to me.  @tekstrand how's it look to you?

tekstrand:
> just in case RAK changes their docs

> @tekstrand how's it look to you?

If RAK changes their docs it's because they have a new way to do this for their hardware. I'm still in the camp that linking out to their site and saying to follow those instructions is the right choice here.

All else looks fine

eresonance:
Hey folks,
I appreciate you have standards for your documentation, which is great to have. However the argument kinda goes against the existing documentation for this particular page as it lists a bunch of steps for programming the bootloader which are similar to, but not exactly the same, as the ones on the RAK website.

For comparison:
https://meshtastic.org/docs/getting-started/flashing-firmware/nrf52/convert-rak4631r/
and
https://docs.rakwireless.com/product-categories/wisblock/rak4631-r/dfu/#converting-rak4631-r-to-rak4631

I'm definitely not an expert in any of this so I'm going to close this issue and replace it with a ticket that someone who knows what they are talking about can tackle:
https://github.com/meshtastic/meshtastic/issues/1664

PR #1655, opened 2025-01-04
Update local-groups.mdx

fromthebeanbag:
Added a few more local groups for Australia

I Added the australia wide and victoria facebook groups 

Provided additional local chapters/groups for Australia

## Why did you change it

Allow those new to meshtastic to find a local group quicker in the documentation website.

### DIFF
--- a/docs/community/local-groups.mdx
+++ b/docs/community/local-groups.mdx
@@ -22,6 +22,8 @@ us on [Discord](https://discord.com/invite/ktMAKGBnBs) to add your group.
 
 ## Australia
 
+- [Meshtastic Australia Wide Community - Facebook](https://www.facebook.com/groups/1169993994163108)
+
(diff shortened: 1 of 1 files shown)


fromthebeanbag:
i'm new to this whole management via github thing, so hopefully i've done this right?

tekstrand:
Hey @fromthebeanbag thanks for the PR - you did things right in GitHub.

Unfortunately we won't be able to merge this in quite yet as both of those groups need their names updated and one of them needs the logo changed. Right now they come off as official meshtastic entities, so they need to be named something like "Meshtastic Community x" and will also need to use their own logo design or they can use the M Powered logo available here https://github.com/meshtastic/design/

fromthebeanbag:
Well as i don't control those groups, i doubt it will happen.. so i may as well kill off this merge request.

PR #1658, opened 2025-01-05
Update usage.mdx

blubdidup:
## What did you change

- existing documentation stays as is

- I only added stuff

## Why did you change it

- added information that should make it easier for first time command line interface (CLI) users to understand how to set parameters
- and slight readability improvements on some of the existing documentation on the page (like added a line break, made some "commands" more visible)

### DIFF
--- a/docs/software/python-cli/usage.mdx
+++ b/docs/software/python-cli/usage.mdx
@@ -12,9 +12,50 @@ This section covers using the "meshtastic" command line executable, which displa
 The `meshtastic` command is not run within python but is a script run from your operating system shell prompt. When you type "meshtastic" and the prompt is unable to find the command in Windows, check that the python "scripts" directory [is in your path](https://datatofish.com/add-python-to-windows-path).
 :::
 
+## Viewing the build-in meshtastic help
+
+By using the below command you get further help directly in the terminal/on the command line.
+
+```shell
+meshtastic --help
+```
+```shell title="A snippet of what you would see in the Resulting Output"
+# A snippet of what you would see in the Resulting Output:
+
+usage: meshtastic [-h | --version | --support] [--port [PORT] | --host [HOST] | --ble [BLE]] [--ble-scan] [--dest !xxxxxxxx] [--ch-index INDEX]
+                  [--configure CONFIGURE] [--export-config] [--get FIELD] [--set FIELD VALUE] [--begin-edit] [--commit-edit] [--get-canned-message]
+                  [--set-canned-message SET_CANNED_MESSAGE] [--get-ringtone] [--set-ringtone RINGTONE] [--ch-vlongslow] [--ch-longslow] [--ch-longfast]
+                  [--ch-medslow] [--ch-medfast] [--ch-shortslow] [--ch-shortfast] [--set-owner SET_OWNER] [--set-owner-short SET_OWNER_SHORT] [--set-ham SET_HAM]
+                  [--seturl SETURL] [--setalt SETALT] [--setlat SETLAT] [--setlon SETLON] [--remove-position] [--pos-fields [POS_FIELDS ...]] [--ch-add CH_ADD]
+                  [--ch-del] [--ch-set FIELD VALUE] [--channel-fetch-attempts ATTEMPTS] [--qr] [--qr-all] [--ch-enable] [--ch-disable] [--info] [--nodes]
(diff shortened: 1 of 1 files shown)


docs/software/python-cli/usage.mdx
```diff
@@ -29,19 +70,141 @@ INFO  | 18:38:04 711 [DeviceTelemetryModule] Sending packet to phone
 INFO  | 18:38:04 711 Telling client we have new packets 28
 ```
 
+It is worth noting that as long as the `meshtastic --noproto` is running in a terminal, you you might not be able to successfully "get" or "set" any device parameters as described in the following chapters.
```
tekstrand, line 73: typo: `you you`

I think we can simplify this whole thing

> While `meshtastic --noproto` is running, you cannot `get` or `set` any device parameters as described in the following chapters.

docs/software/python-cli/usage.mdx
```diff
@@ -29,19 +70,141 @@ INFO  | 18:38:04 711 [DeviceTelemetryModule] Sending packet to phone
...
-You can get a list of user preferences by running '--get' with an invalid attribute such as 'all'.
+You can get a list of user preferences by running `--get` with an invalid (not known to meshtastic) attribute such as `all`.
 
 ```shell
 meshtastic --get all
 ```
 
-## Changing settings
+```shell title="A snippet of what you would see in the Resulting Output"
+# A snippet of what you would see in the Resulting Output:
```
tekstrand, line 84: Duplicate content from the title

docs/software/python-cli/usage.mdx
```diff
@@ -29,19 +70,141 @@ INFO  | 18:38:04 711 [DeviceTelemetryModule] Sending packet to phone
...
+    display.heading_bold
+    display.oled
+    display.screen_on_secs
+    display.units
+    display.wake_on_tap_or_motion
+
+...
+```
+
+## Getting a list of User Preferences including the Values they are currently set to
```
tekstrand, line 122: This title is too long for an h2, h2's show up in the navbar on the side.

> Getting Preferences and Values

Should suffice

docs/software/python-cli/usage.mdx
```diff
@@ -29,19 +70,141 @@ INFO  | 18:38:04 711 [DeviceTelemetryModule] Sending packet to phone
...
 
-You can also use this tool to set any of the device parameters which are stored in persistent storage. For instance, here's how to set the device
-to keep the Bluetooth link alive for eight hours (any usage of the Bluetooth protocol from your phone will reset this timer)
+You can get this by running `--info` which reads and displays the radio config information. Which will also include a list of all "Nodes in mesh" known to the device.
```
tekstrand, line 124: Strange wording, especially in the second sentence. Not sure we need this content at all as the command and output below display everything

tekstrand (changes requested):
Halfway done with review, needs some work. I'll look again later when I have time. Thanks for the updates!

docs/software/python-cli/usage.mdx
```diff
@@ -29,19 +70,141 @@ INFO  | 18:38:04 711 [DeviceTelemetryModule] Sending packet to phone
...
+```
+will result in the message:
+
+```shell title="Error Message"
+...
+LocalConfig and LocalModuleConfig do not have attribute autoScreenCarouselSecs.
+...
+```
+We will need to add the `display` part to our command by adding it before the `autoScreenCarouselSecs`. And we will use a `.` as seperator between them.
+So the command would be:
```
tekstrand, line 194: We don't need to showcase the wrong way to do it, we can just show the right way.

docs/software/python-cli/usage.mdx
```diff
@@ -29,19 +70,141 @@ INFO  | 18:38:04 711 [DeviceTelemetryModule] Sending packet to phone
...
+```shell title="Expected Result"
+Connected to radio
+Set display.auto_screen_carousel_secs to 20
+Writing modified preferences to device
+```
+
+### More Examples
+
+For instance, here's how to set the device
+to keep the Bluetooth link alive for eight hours (any usage of the Bluetooth protocol from your phone will reset this timer)
```
tekstrand, line 207:
```suggestion
Keeping the Bluetooth link alive for 80 hours. Note that any usage of the Bluetooth protocol from your phone will reset this timer
```

This can be simplified. Also, is `phone` right here? Should we just say client device?

tekstrand (commented):
Alright, all finished with the first round of review. Thanks again for this, I think these are going to be helpful improvements.

PR #1662, opened 2025-01-07
[docs] Add gps toggle button info for LilyGo

johnnyasantoss:
Just completing the docs


johnnyasantoss:
I've just didn't update tbeam because it has images and the text refers to them.

PR #1669, opened 2025-01-10
Linux docs overhaul

vidplace7:
WIP!

## What did you change
Overhaul Linux docs to include new Debian / Raspbian / Fedora / Docker packaging.

Awaiting:
- https://github.com/meshtastic/firmware/pull/5791
- https://github.com/meshtastic/firmware/pull/5735

PR #1670, opened 2025-01-10
Add Icons to Tabs (where it makes sense)

rcarteraz:
## What did you change
Add icons to the tab items

## Why did you change it
@vidplace7's upcoming PR (#1669) makes use of icons in the tabs for the specific OSs which looks great. So I decided to add icons to other tabs (where it makes sense). 

### After
Screenshot 2025-01-10 at 10 53 17 AM
Screenshot 2025-01-10 at 10 53 34 AM
Screenshot 2025-01-10 at 10 53 23 AM


pdxlocations (approved):
I don't suppose there's a way to make this code a reusable block  so it doesn't have to be added to every page?

PR #1686, opened 2025-01-23
Update building.mdx

Xaositek:
Added a section for "Update the Build Environment"

## What did you change
Added a new section for "Update the Build Environment"

## Why did you change it
I figured if I had the question, someone else would too, just trying to minimize friction for engagement.


Xaositek:
Updated per @rcarteraz comments

PR #1690, opened 2025-01-28
Update trademark-grants.mdx

dotspencer:
## What did you change

Add Smith Off-Grid to trademark grants


garthvh:
There is no need for a trademark grant to use the m powered logo, it can be revoked if used improperly.

PR #1716, opened 2025-02-14
Local Group Submission Guidlines

pdxlocations:
This PR adds an abbreviated list of guidelines for groups wishing to list on the Community Groups page.

Far too many PR's for addition are submitted that don't follow the trademark guidelines so I'm submitting this as a draft for consideration. Feel free to adjust the content to best align with project goals.

I had considered only linking to the trademark page, but I worry a many users would not take the time to read more than a few sentences.  I believe a short summary could improve our chances of getting the message out. 

I'm not set on using an admonition here, though I do feel it helps catch the eye.

### DIFF
--- a/docs/community/local-groups.mdx
+++ b/docs/community/local-groups.mdx
@@ -16,10 +16,10 @@ them for assistance in getting started or if you're interested in contributing t
 organizer with an online presence and wish to be included in this list, please edit this page directly or reach out to
 us on [Discord](https://discord.com/invite/ktMAKGBnBs) to add your group.
 
-:::info Important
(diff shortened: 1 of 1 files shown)


tekstrand (approved):
LGTM.

Do we have any requirements around only allowing one link per group, or only allowing the group in one place(e.g. a group that spans multiple countries or states)?

pdxlocations:
> Do we have any requirements around only allowing one link per group, or only allowing the group in one place(e.g. a group that spans multiple countries or states)?

That's an interesting thought as it does come up from time to time.  We could add a note or just address it in the PR as it's kind of rare.  I already think this list (I wrote) is a bit dry and making it much longer may discourage engagement.?

PR #1721, opened 2025-02-17

battlehax:
Update pi compatibility list.

## What did you change
Removed Pi Zero from compatibility list.

## Why did you change it
The new installation instructions and repo are incompatible with the cpu architecture of the pi 1/zero/zero w and will lead to a a segfault if attempted.


vidplace7 (approved):
Unfortunately OpenSUSE Build Service doesn't currently support building `armv6l` Raspbian packages (only `armv7l`). Good catch :+1: 

I'm filing a feature request with OpenSUSE to add support for this, but also not holding my breath :sweat_smile:

vidplace7:
https://github.com/meshtastic/firmware/pull/6104

PR #1725, opened 2025-02-20
Add disable triple click

noon92:
Was missing


pdxlocations (approved):
--get device.disable_triple_click shows False after Factory Reset, so let's go with False!

PR #1728, opened 2025-02-23
Linux: Add RedHat EPEL Install instructions

vidplace7:
Adds docs for installing meshtasticd on RedHat derivatives.

This encompasses CentOS Stream, Redhat Enterprise Linux, AlmaLinux, Rocky Linux, and Oracle Linux (and probably others).

image


vidplace7:
Beta builds are already available in the meshtastic COPR so this should be good to merge :+1:

pdxlocations (approved):

PR #1745, opened 2025-03-12
Update Python Development Instructions

pdxlocations:
The requirements.txt file is no longer in the repo. Based on a comment from @ianmcorvidae on [discord](https://discord.com/channels/867578229534359593/871553765105348668/1347980033247150110), I've replaced it with
`poetry install --all-extras --with dev,powermon`. Marking as draft since I don't have working knowledge of the process.

Before:
Screenshot 2025-03-12 at 2 31 25 PM

After:
Screenshot 2025-03-12 at 2 31 49 PM


ianmcorvidae (approved):
Yeah, this is a sensible change. I'm not sure how accurate the rest of the page is, but that can be a problem for another time/another PR.

Issue #1567, opened 2024-11-11
Documentation Change Request: Include pipx installation for CLI and esptool


pdxlocations:
I believe it would be useful to include the recommendation and install for pipx on pages where we talk about pip.

https://meshtastic.org/docs/software/python/cli/installation/
https://meshtastic.org/docs/getting-started/flashing-firmware/esp32/cli-script/

PR #1751, opened 2025-03-15
Update mac cli install to use pipx

pdxlocations:
This PR updates the mac os CLI installation instructions to use pipx
### Before

Screenshot 2025-03-14 at 7 39 06 PM

### After
Screenshot 2025-03-14 at 7 38 48 PM


pdxlocations:
helps with #1567 for mac at least.

PR #1752, opened 2025-03-15
add InkHUD

todd-herbert:
This PR adds initial documentation describing the alternative "InkHUD" UI. Please let me know if there are any changes that could be made. Also happy to push the design assets somewhere, in case anyone else ever needs to update the images


docs/software/inkhud/applets.mdx
```diff
@@ -0,0 +1,138 @@
+---
+id: applets
+title: InkHUD Applets
```
tekstrand, line 3: This might be a discussion for somewhere else, but why is this called Applets? Just wondering if the layperson is going to understand that here in the docs

docs/software/inkhud/index.mdx
```diff
@@ -0,0 +1,148 @@
...
+  />
+</div>
+
+:::tip
+Only wanted a quick peek at the clock? Long-press a second time to close the menu. 
+:::
+
+### Options
+
+#### Applet Selection
```
tekstrand, line 84: @pdxlocations @rcarteraz want your thoughts here

Not sure if using h4's is a pattern we're following already and this just sticks to that, but having each one of these as an h4 seems not quite right. I think that using a table or just bullet points could be better.

What are y'alls thoughts?

docs/software/inkhud/applets.mdx
```diff
@@ -0,0 +1,138 @@
...
+description: A description of the various applets available as part of the InkHUD user interface.
+---
+
+This list of applets is expected to grow as InkHUD continues to evolve.
+
+## All Messages
+
+Displays the most recent message. Both in-channel messages and DMs will be shown.
+
+<div align="center">
```
tekstrand, line 16: Was centering all images here just following a paradigm from another page? IMO (which is really just an opinion) it makes it harder to read the page, and these would be better left aligned.

tekstrand (changes requested):
This is superb! Thank you for taking the time to write this up. Just a few comments and changes I am suggesting let me know if I can be of help

docs/software/inkhud/applets.mdx
tekstrand: @rcarteraz what was the reasoning behind it for the MUI page?

docs/software/inkhud/index.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+If enabled, show a small icon which indicates battery's state of charge, in the top-right corner of the display.
+
+## Multiple Applets
+
+InkHUD can show multiple applets at the same time; a feature knows as _tiling_. To change the amount of tiles, select `Options > Layout` in the on-screen menu.
+
+### Moving between tiles
+
+<Tabs
```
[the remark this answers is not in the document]
todd-herbert: I did initially have it with two separate headings, but I personally found it quite distracting to have the two animated elements next to each other.

https://github.com/user-attachments/assets/ecf8fcfa-6c9e-43f0-b64a-3a9586d97472

Would you guys agree that it's a bit jarring? Is there maybe a better way to separate these two elements than with tabs?

docs/software/inkhud/applets.mdx
todd-herbert: The image centering was definitely following the MUI page, although I did experiment with left alignment too and found that I personally preferred the centering. I'm not attached to that though, so happy to go with whatever the consensus ends up being.

docs/software/inkhud/applets.mdx
todd-herbert: It's a valid point. I remembering deliberating over that term for quite a while and settled on applet as a reasonable description of the role that the `InkHUD::Applet` component plays within the firmware. The feedback is useful insight, because I hadn't considered just how vague the term sounds outside the context of the code, and this point am so overexposed to it that it's hard to imagine how it might sound to a user encountering it for the first time.

I'm open to the idea of changing the terminology, certainly for user-facing stuff like the doc and on-screen menu. Would you guys have any possible suggestions for a term that might be more appropriate? I remember it being a little bit tricky to find a term which was unambiguous, with words like "screen" having several possible meanings.

docs/software/inkhud/applets.mdx
tekstrand: If we're going to center everything, I think that we should just do it with css instead of adding a bunch of html elements that we have to manage later. The `margin: auto` for `markdown img` is already to to left/right auto, the only thing missing is setting img to `display: block`.

docs/software/inkhud/index.mdx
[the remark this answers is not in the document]
tekstrand: I don't find it distracting, personally, but I understand your point. I'd like to heir on the side of simplicity by just using headings, but I also won't die on this hill!

docs/software/inkhud/applets.mdx
tekstrand: No argument on the technical correctness of the term, however I don't find that it is in common use. If we were to do some user testing compared to terms such as Widget or Module, I believe we would see easier comprehension.

docs/software/inkhud/applets.mdx
tekstrand: All good, here's the fix https://github.com/meshtastic/meshtastic/pull/1757

docs/software/inkhud/applets.mdx
tekstrand: But, that is when I am bringing it up, right? It's still in development, pre alpha, and it isn't documented yet.

I get that this PR might not be the right place, happy to bring it up wherever you feel is the right place.

docs/software/inkhud/applets.mdx
tekstrand: @todd-herbert that change has been merged. You should be able to go through and remove any of the html around the `img` tag for the centering stuff and it'll still be centered automatically by css 🥳

docs/software/inkhud/applets.mdx
tekstrand: I went ahead and  raised the conversation in the contributor lounge channel https://discord.com/channels/867578229534359593/931765413350613052/1351654426435387543

PR #1755, opened 2025-03-17
Note how to install the web client via CLI

interfect:
## What did you change

I added some documentation for flashing the web client via CLI.

## Why did you change it

I tried to install the web client on my device with `./device-install.sh -f littlefswebui-2.5.20.4c97351.bin`, which is very wrong but attempts to work anyway.

Also, if you're not using Google Chrome, you won't have the web flasher available with the checkbox to install the web UI, and you also probably want the web UI hosted off the device, since you won't have anything but the network method of talking to the device from a browser, and connecting the cloud-hosted web UI to the device over the network at a different origin requires a special SSL certificate acceptance dance. So anyone not on Chrome is going to be using the CLI flasher *and* probably will want the on-device web UI.

### Before
image

### After
image

### DIFF
--- a/docs/getting-started/flashing-firmware/esp32/cli-script.mdx
+++ b/docs/getting-started/flashing-firmware/esp32/cli-script.mdx
@@ -348,6 +348,8 @@ device-update.bat -f firmware-BOARD-VERSION-update.bin
   </TabItem>
 </Tabs>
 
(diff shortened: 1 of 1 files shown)


pdxlocations:
Nice!  I've tested this and can confirm it works.  @interfect would you sign the CLA for further review?

PR #1763, opened 2025-03-21
Update lora.mdx

jschrempp:
Added note that for Transmit Power the Apple apps do not show 0 as an option and that the user should pick the largest value.


todd2982:
Will close #1764

PR #1768, opened 2025-03-22
Update key restore command

todd2982:
Changed --config to --configure to match python module help text.  This matches the current 2.6.0 python module.

image


pdxlocations (approved):
while --config does work since the parser allows abbreviations, I agree with this change.  There could be a `config-xxx` in the future which would make this abbreviation fail.

PR #1774, opened 2025-03-28
Add SouthCoast Mesh to Local Groups -> United States -> Massachusetts

jhstatewide:
## What did you change

Adds our local user group to the local groups page.

## Why did you change it
We changed it because we wanted to appear in the official list of groups


pdxlocations (approved):
Looks great!

PR #1784, opened 2025-04-02
Add Default Slot to Frequency Slot Calculator

rcarteraz:
## What did you change
Modified the frequency slot calculator so that it will also show the default frequency slot for the selected modem_preset and region. * This is based on the channel hashing only of the modem_preset. 

## Why did you change it
Useful information to have and this way you don't have to either run `--noproto` or look at the app. Just select the modem_preset and channel and it'll display it. 

### Before
Screenshot 2025-04-02 at 1 19 20 PM

### After
Screenshot 2025-04-02 at 1 19 36 PM

### DIFF
--- a/docs/software/meshtasticui/index.mdx
+++ b/docs/software/meshtasticui/index.mdx
@@ -75,6 +75,10 @@ Certain supported devices do not have automatic firmware builds or web flasher s
 
 ## Usage
 
+### SD Card Support
+
+For devices with supported SD card slots, ensure the card is properly configured before use. The SD card should be formatted using MBR. MUI supports both FAT32 and exFAT partitions, with exFAT being the recommended format.
+
 ### Initial Setup
 
 Upon first booting MUI on a device or after factory resetting a device, the initial screen displayed is a quick configuration screen to set the device's region and both long name and short name. This screen allows these settings to be set at the same time, requiring only one reboot, rather than separately setting each one and rebooting twice.


thebentern (approved):

src/components/tools/FrequencyCalculator.tsx
```diff
@@ -289,6 +289,33 @@ const modemPresets = new Map<
...
 ]);
 
+// Helper function to format the channel name (e.g., LONG_FAST -> LongFast)
+const formatChannelName = (name: string): string => {
+  return name
+    .toLowerCase()
+    .split("_")
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join("");
+};
```
ianmcorvidae, line 299: This won't be right. Many presets don't use something following this simple transformation.

The actual values are all in the switch statement at https://github.com/meshtastic/firmware/blob/master/src/DisplayFormatters.cpp#L5

src/components/tools/FrequencyCalculator.tsx
```diff
@@ -326,12 +354,22 @@ export const FrequencyCalculator = (): JSX.Element => {
...
+    setDefaultSlot(defaultSlot);
   }, [modemPreset, region, channel]);
 
   return (
-    <div className="flex flex-col border-l-[5px] shadow-md my-4 border-accent rounded-lg p-4 bg-secondary gap-2">
-      <div className="flex gap-2">
-        <label htmlFor="modemPreset">Modem Preset:</label>
+    <div class="flex flex-col border-l-[5px] shadow-md my-4 border-accent rounded-lg p-4 bg-secondary gap-2">
+      <div class="flex gap-2">
+        <label for="modemPreset">Modem Preset:</label>
```
ianmcorvidae, line 372: These changes from `className` and `htmlFor` to the HTML equivalents won't work right in JSX, most likely. I could be wrong about that though.

src/components/tools/FrequencyCalculator.tsx
ianmcorvidae: Okay, maybe not "many", looking again. But two of them, anyway, haha.

PR #1785, opened 2025-04-02
Fix FreqSlot Calc

rcarteraz:
## What did you change
The formatting of the channel name to do the hash was wrong, 
also getting rid of very long slow since it's deprecated. 

## Why did you change it
Because copilot failed me and @ianmcorvidae called me out. I figured it should. 

I've tested this manually with a heltec v3 setting random combinations and verified that the frequency is still correct, as well as doing random modem_preset and regions and they seem to be matching what the firmware does based on this testing. 

Is it the right/proper way to do this? Most assuredly not. But does it work? It would seem it does. Someone actually experienced and smarter than me (or rather copilot 😅) is welcome to PR a replacement.

### DIFF
--- a/src/components/tools/FrequencyCalculator.tsx
+++ b/src/components/tools/FrequencyCalculator.tsx
@@ -1,4 +1,4 @@
-import { Protobuf, Types } from "@meshtastic/js";
+import { Protobuf, type Types } from "@meshtastic/js";
 import { useEffect, useState } from "react";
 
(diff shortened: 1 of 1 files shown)


ianmcorvidae (approved):
I think this is good, certainly as far as the name it is. I'm not rapidly parsing all the changes later on in the file but nothing jumps out at me, at least.

PR #1786, opened 2025-04-03
Update web-client.mdx

Will-wastelander:
Adding SBC Linux update process.

## What did you change
Add SBC Linux manual update section.

## Why did you change it

Because there is no process provided to update SBC Linux WebUI.


vidplace7:
Related to https://github.com/meshtastic/web/pull/559

We should not be instructing users to upgrade the web portion manually. This has high potential to result in broken package upgrades in the future, and will result in additional support burden on the Linux maintainers.

I'd much rather direct users to our Daily or Alpha builds if they'd like to test the latest and greatest web releases.

PR #1787, opened 2025-04-05
Fix external links

bobidle:
## What did you change
Change some external links (Redirects or Not Found issues)

## Why did you change it

1. Redirects could get outdated and result in a Not Found error
2. Not Found issues (404) should be fixed

Please check the `etsy.com` and `printables.com` changes again. I am not sure if the addresses changed because of i am access these links from a different location.

`https://garthvh.com/store` is not available for me, i assumed liking to the `etsy.com` shop would be a good replacement, but i am unsure about it.


docs/legal/trademark-grants.mdx
garthvh, line 1: Just https://garthvh.com should work

garthvh (changes requested):

docs/legal/trademark-grants.mdx
garthvh: That is fine

garthvh:
> For the future, let's not include changes to the legal section with other changes and keep those separate. Thanks!

There is a comment that needs to be resolved be fore merging this, also need to check with @tropho23 as his link changed too and probably should not be the direct store link.

tropho23:
> > > For the future, let's not include changes to the legal section with other changes and keep those separate. Thanks!
> > 
> > 
> > There is a comment that needs to be resolved be fore merging this, also need to check with @tropho23 as his link changed too and probably should not be the direct store link.
> 
> It can’t be merged until you either approve the changes or your review is dismissed. I have done neither so it’ll sit until you’re ready for that.
> 
> As for Tony’s link they’re essentially the same, different formats but same destination but he’s welcome to chime in if he prefers one over the other.

Yes, I changed link names based on slight adjustments to case name to match preferred naming convention, e.g. "T-Beam case for Meshtastic" vs "Meshtastic T-Beam case".  Also specifying T-Beam v1.x after I began offering a T-Beam Supreme case as well, etc.

I'm good with those changes for my content's link as suggested in the PR.

simon-muzi:
While we're in here, can we update my link as well?
(wasn't sure how to add it as a change request)

- - Grant: [Simon - muzi.works](https://instagram.com/muzi.works)
+ - Grant: [Simon - MUZI WORKS LLC](https://muzi.works)

PR #1793, opened 2025-04-08
docs: Telegram link for Meshtastic enthusiasts in Colombia

hpsaturn:
Added link to the Colombian Meshtastic enthusiast Telegram group to help local community members connect and collaborate.


hpsaturn:
I don't understand you, I can see other groups that already have a similar link that "mestasticxx" in the Telegram group, for instance, the Teck group for Germany. Why we need change it and them not? We can't change the link of the group. What happen here?

PR #1822, opened 2025-04-26
Update radio-settings.mdx

jschrempp:
per discussion in discord, Channel is the wrong term to use here.

## What did you change
Column header of the table used the term Channel which is not correct. 

## Why did you change it
To avoid confusion with the Meshtastic term Channel for group communication logical meshes.

### Before
Screenshot 2025-04-25 at 9 14 47 PM


garthvh (changes requested):
Can you capitalize radio? Looks good otherwise

jschrempp:
Made the change requested by @garthvh - good point, thanks.

garthvh (approved):

PR #1835, opened 2025-05-04
Update mesh-alg.mdx Reworded about rebroadcasting

jschrempp:
## What did you change
A few words in a sentence on rebroadcasting.

## Why did you change it
Original sentence was awkward. I tried to improve it for readability without changing the meaning.

### Before
However, an important difference in Meshtastic is that before rebroadcasting, a node listens a short while to see if no other node already rebroadcasted it.

### After
However, an important difference in Meshtastic is that before rebroadcasting, a node listens a short while to see if another node has already rebroadcasted the packet.


tekstrand (approved):
Nice improvement.

PR #1837, opened 2025-05-04
Update index.mdx Changed Yet to Thus

jschrempp:
## What did you change
Changed one word - "yet" to "thus"

## Why did you change it
In reading the sentence, I think "yet" is confusing. I believe the writer's intended meaning is more clear using "thus". If I have misunderstood the intent of the sentence, please reject this PR.

### Before
Other protocols, such as _misconfigured TLS1.2_, are far more impacted by Harvest now, Decrypt later attacks because their keys will be broken by near-future Quantum-Computers, YET they do not provide PFS.

### After
Other protocols, such as _misconfigured TLS1.2_, are far more impacted by Harvest now, Decrypt later attacks because their keys will be broken by near-future Quantum-Computers, THUS they do not provide PFS.

### DIFF
--- a/docs/about/overview/encryption/index.mdx
+++ b/docs/about/overview/encryption/index.mdx
@@ -35,7 +35,7 @@ This means Meshtastic is vulnerable to [**« Harvest now, Decrypt later »**](ht
 - One of your nodes being stolen.
 - Any unknown bug revealing the channel key to an attacker.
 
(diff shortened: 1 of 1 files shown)


jschrempp:
That is a different read than I had. I can appreciate this. The idea is that Meshtastic is no worse in PFS than these other big-name software libs. How about this:

"Other protocols, such as misconfigured TLS1.2, are far more impacted by Harvest now, Decrypt later attacks because their keys will be broken by near-future Quantum-Computers, yet they also do not provide PFS."

PR #1848, opened 2025-05-08
Victoria, Avustralia Meshtastic group is added to local-groups.mdx

c0d3rk1d:
## What did you change
Victoria, Avustralia Meshtastic group is added to local-groups.mdx


tekstrand (changes requested):
The logo and group name will have to be changed to align with the guidelines before we merge this

PR #1849, opened 2025-05-08
South Africa is added to local-groups.mdx

c0d3rk1d:
## What did you change
South Africa's Discord and MeshInfo pages are added.


c0d3rk1d:
@rcarteraz No, I'm not. I was looking at other communities to see what they’re doing, and I added some of them to this list. If that's the case, please reject all my PRs, and I'll stop adding to the list and look for something else to improve in Meshtastic. ;)

PR #1853, opened 2025-05-10
Update telemetry.mdx

NomDeTom:
Add AHT10 and AHT20 to supported sensor list


tekstrand (changes requested):
This is going to conflict with this other PR which seems more complete https://github.com/meshtastic/meshtastic/pull/1857/files

tekstrand:
@rcarteraz merging this and the other one caused duplicates on the page https://meshtastic.org/docs/configuration/module/telemetry/

PR #1858, opened 2025-05-13
Replaced twitter logo with X.com

RCGV1:
## What did you change

Replaced the twitter logo and name with x.com branding and website.
## Why did you change it

To be consistent with the twitter rebrand.
### After
Screenshot 2025-05-13 at 11 27 36 AM


src/pages/index.tsx
```diff
@@ -28,13 +28,13 @@ const SocialCards: SocialCardProps[] = [
     ),
   },
   {
-    color: "bg-[#4A99E9]",
-    link: "https://twitter.com/TheMeshtastic",
+    color: "bg-[##000000]",
```
tekstrand, line 31: Looks like you have an extra #

src/pages/index.tsx
```diff
@@ -28,13 +28,13 @@ const SocialCards: SocialCardProps[] = [
...
+    link: "https://x.com/TheMeshtastic",
     children: (
       <img
-        alt="twitter"
+        alt="x.com"
         className="m-auto h-10"
-        src="/img/homepage/Twitter-Logo-White.svg"
+        src="/img/homepage/X.com-White-Logo.svg"
```
tekstrand, line 37: We're not using the old image anywhere so we should remove it from the repo

tekstrand (changes requested):
Thanks for the update! A couple small changes

tekstrand (approved):

PR #1865, opened 2025-05-26
Add Chile US frequency as the official

raztor:
## Included Chile in the list of countries with it's respective frequency zone (US) and url to national RF regulation

This fork adds Chile to the list in LoRa Region by country

## Why did you change it
We are recently planning and creating the network for Chile and we need to appear on the official wiki so the new users know what frequency to use.

We are also building our wiki (wiki.meshtastic.cl and we have our own mqtt server mqtt.meshtastic.cl)

### Before

image

### After
image

### DIFF
--- a/docs/community/local-groups.mdx
+++ b/docs/community/local-groups.mdx
@@ -23,6 +23,9 @@ To be listed here, your group must be in compliance with our [Trademark Guidelin
 - Server invitation links must be set to never expire.
 :::
 
+## Chile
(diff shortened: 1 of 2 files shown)


raztor:
I forgot to add the group to local groups section. I commited the new changes

raztor:
@rcarteraz Sorry I forgot about that.

I will make a new PR with a new domain that follows the guidelines.

Thx

PR #1869, opened 2025-05-30
Update usage.mdx

NomDeTom:
Remove link to dead site, hopefully linking to something suitable instead.

(as reported by discord user Forestpas3)


pdxlocations:
This looks like a suitable link, but seems much more confusing than the old one.  I wonder if linking to external sites for guidance that can as easily be googled is the best idea?

PR #1871, opened 2025-05-30
Update region-by-country.mdx

erayd:
Add a note to reduce confusion for new users.


pdxlocations:
My thought is that this note doesn't fit within the context of this page and it would best be left to regional social outlets to discuss preferred bands.  Just my thoughts of course, always open to discussion.

rcarteraz:
Agree with @pdxlocations

PR #1872, opened 2025-05-31
Update g1 station to indicate unsupported status

Hnikar-az:
station-g1 is not located in bin/device-install.sh or in boards list on https://github.com/meshtastic/firmware/tree/9799f10e6363703bf64490044dbac813f0ddf2ea/boards changed the language on the page to indicate it is no longer supported.

## What did you change

"The Station G1 remains a supported device." changed to The Station G1 is no longer a supported device.
## Why did you change it

started with this rabbit hole seeking a solution to the problem of installing the most recent firmware on the station-g1: https://github.com/meshtastic/firmware/issues/6929 and realized that the station G1 is not listed on the boards or in the device-install.sh script file

### DIFF
--- a/docs/hardware/devices/b-and-q-consulting/station-series/index.mdx
+++ b/docs/hardware/devices/b-and-q-consulting/station-series/index.mdx
@@ -25,7 +25,7 @@ values={[
 
 :::info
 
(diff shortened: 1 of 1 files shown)


Hnikar-az:
I'm wrong, it is still supported by the web flasher but maybe not the cli

PR #1877, opened 2025-06-05
Update region-by-country.mdx

jiehoes:
Add Indonesia Regulation document and LoRa Region


rcarteraz:
We do not have a dedicated Indonesia region. That would need to be added to the firmware before this could be added to the docs.

PR #1879, opened 2025-06-05
Make frequency slot calc use default slot instead of 0

rfschmid:
## What did you change
Make Radio Settings doc's Frequency Slot Calculator tool set the "Frequency Slot" to the value of the "Default Frequency Slot" when changing Modem Preset or Region, rather than the first (1).

## Why did you change it
The default frequency slot is the one that will be used most often, so it makes more sense to initialize the frequency slot to the default rather than making the user manually select the default frequency slot from the list. This lead to a bit of confusion in this discord chat:
https://discord.com/channels/1215670460264742922/1379994081744257215/1380265051059130379

### DIFF
--- a/src/components/tools/FrequencyCalculator.tsx
+++ b/src/components/tools/FrequencyCalculator.tsx
@@ -352,16 +352,6 @@ export const FrequencyCalculator = (): JSX.Element => {
 
       setNumChannels(calculatedNumChannels);
 
-      // Reset the channel to 0 when modemPreset or region changes
-      const defaultChannel = 0;
-      setChannel(defaultChannel);
-
-      // Recalculate the frequency for the default channel
-      const newChannelFrequency =
-        selectedRegion.freqStart +
-        selectedModemPreset.bw / 2000 +
-        defaultChannel * (selectedModemPreset.bw / 1000);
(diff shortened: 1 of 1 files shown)


rcarteraz:
The default frequency slot is already shown, and making a change because of one person's confusion isn't ideal. This has been available for quite bit of time with no complaints. Should this continue to be an issue, we'll revisit.

rfschmid:
@rcarteraz 

Thanks for the review!

I understand it's been like this a while but making the default the default certainly makes intuitive sense to me personally. I don't think anyone is confused, just not paying close attention and assuming the default would default to the default, and a second person agreeing it was more intuitive to make the default the default 🙂

If we can't do this, as an alternative would there be any appetite for adding an additional field to show the frequency for the default slot alongside the display for the value of the default slot? It seems like unnecessary work for users to have to scroll and select the frequency slot manually if they're just interested in seeing the default. In many cases the list of frequency slots is very long!

rfschmid:
@rcarteraz just wanted to check in and see if you'd considered my question above about adding an additional field for default slot frequency if there's no appetite to change the default of the frequency input. Let me know what you think!

PR #1887, opened 2025-06-11
Added Chile to wiki

raztor:
## What did you change
Added Chile to region by frequency with it's respective RF regulation plan

Add chile to the list
## Why did you change it
We created a community of meshtastic in Chile. We want to be part of the wiki


raztor:
I changed the domain from meshtastic.cl to meshchile.cl to follow brand guidelines because of the previous rejected PR

tekstrand:
meshchile.cl doesn't resolve for me

raztor:
@tekstrand for the root domain there will be a redirection on a couple hours.

But the url I setup on the wiki is links.meshchile.cl

PR #1889, opened 2025-06-12
Update TEcho Description

pdxlocations:
It is not the newest product from LILYGO


rcarteraz:
> It is not the newest product from LILYGO

Too funny

PR #1904, opened 2025-06-25
Update range-test.mdx on message visability and that gps is optional.

willhubs:
Added that only nodes that have the range test module enabled will see the range test messages.

Also added that gps on the node/phone is optional.

### DIFF
--- a/docs/community/local-groups.mdx
+++ b/docs/community/local-groups.mdx
@@ -147,6 +147,7 @@ To be listed here, your group must be in compliance with our [Trademark Guidelin
 ## Greece
 
 - [Meshtastic Crete Users Group](https://www.facebook.com/groups/meshcrete)
+- [Athens Meshtastic Node Runners](https://www.facebook.com/groups/athensmesh)
 
 ## India
 


docs/configuration/module/range-test.mdx
```diff
@@ -8,10 +8,10 @@ description: Learn to configure the range test module to create reports, maps, a
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 
-This module allows you to test the range of your Meshtastic nodes. It requires two nodes:
+This module allows you to test the range of your Meshtastic nodes. It requires two nodes, only nodes that have the module enabled will see the range test messages:
```
pdxlocations:
```suggestion
This module lets you test the communication range between two Meshtastic nodes. Only nodes with the module enabled will show range test messages on their connected clients.
```

willhubs:
I don't know what the vercel thing is, but it wouldn't let me sign up without some further verification or something.  I'm not planning on doing whatever that is.

pdxlocations:
Don't worry about Vercel, only the maintainers can trigger a preview.

pdxlocations:
I've left a comment on the wording (above) to maybe make it a little more concise.  What do you think?

PR #1912, opened 2025-06-27
Add language dropdown picker to the titlebar

danditomaso:
## What did you change

Using the Docusarus configuration file I used a built in method to add the language dropdown to the right side of the navigation bar.

## Why did you change it

Its important to offer users an option to override their browser language selection. As well I mentioned it to @fifieldt and he wanted to go forwards with the change. 

### Before
image

### After
image
image


rcarteraz:
I'm all for this but the majority of the languages have little to no translations available so it gives the appearance we have translations when in reality we do not. I'm not sure that's the best user experience.

fifieldt:
I read the docusaurus docs and it seems there's no way to use localeDropdown with just a subset of the locales :( So we'll need to do something custom.

rcarteraz:
I think what we should do then is not add that locale as an option unless we have the about, getting started, and configuration sections translated then. So at least the most visited pages have translations.

PR #1914, opened 2025-06-28
Update local-groups.mdx

Skordy:
## What did you change
Removed Tucson Meshtastic Community link from Arizona

## Why did you change it
Discord is no longer maintained, all members have moved to the larger Arizona Meshtastic Community Discord.


ianmcorvidae (approved):
Sadly, yup. Maybe someday we can revive it :(

PR #1916, opened 2025-06-30
Add Indiana to local-groups.mdx

GMart:
## What did you change
Add two groups in Indiana

## Why did you change it
Got approval from Indiana Mesh Discord to post here. The BTown mesh is still being developed. They currently meet on a HAM radio local Discord.


GMart:
@pdxlocations, thanks. Would 'IndianaMesh' or 'Indiana Mesh' be ok as a group name?

pdxlocations:
> @pdxlocations, thanks. Would 'IndianaMesh' or 'Indiana Mesh' be ok as a group name?

Yes, both of those are okay.

pdxlocations (approved):

PR #1918, opened 2025-07-01
Update mqtt.mdx

fifieldt:
per @Code1110 , 
"""This should be reflected in the docs as well. https://meshtastic.org/docs/configuration/module/mqtt/ still says the default is 900 seconds and doesn't mention that the minimum is now 3600 seconds. Entering a value smaller than 3600 seconds in the UI changes the value silently to 3600 seconds and is therefore confusing."""


garthvh (approved):
True

rcarteraz:
I'm pretty sure there's a number of changes for default values that need to be updated with recent throttling changes in the firmware.

PR #1922, opened 2025-07-05
Create index.md for T-Deck PLus

oliverkrystal:
## What did you change
Added file for T-Deck plus, added notes about GPS module connection

## Why did you change it
To try to improve documentation for t-deck plus users

### DIFF
--- a/docs/hardware/devices/lilygo/tdeck+/index.md
+++ b/docs/hardware/devices/lilygo/tdeck+/index.md
@@ -1,0 +1,83 @@
+---
+id: tdeck+
+title: LILYGO® T-Deck Plus
+sidebar_label: T-Deck PLus
+sidebar_position: 6
+---
+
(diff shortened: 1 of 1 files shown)


pdxlocations:
This documentation is nearly identical to the tdeck page.  Could we simply add information about the GPS pins there and keep it to one page?

oliverkrystal:
> This documentation is nearly identical to the tdeck page. Could we simply add information about the GPS pins there and keep it to one page?

Would this be preferred?  I suppose it would make sense.  My other thought was to add a note about the power issues and that 2.7.1 would appear to have corrected it - but that would also apply to any eps32-s3 device.  Let me know if I should submit a new pull request to add these notes to the t-deck page instead.

pdxlocations:
Not sure what the power issues are about, but for now lets keep it concise and just have one TDeck page since they use the same firmware and are essentially the same device.  We may have to revaluate what we do with the TDeck section once the Pro is also supported.

PR #691, opened 2023-06-28
Update trademark-grants.mdx

Yukonus:
Requesting trademark grant for K9 Rocket Technologies

PR #1924, opened 2025-07-05
Remove Bundled WebUI References

pdxlocations:
The webflasher no longer bundles the webUI on esp32 devices.  This PR removes the instructions for installation and usage.


roberthadow:
May I suggest that the superceding approach be tested for robustness.  See Client.meshtastic.org not connecting [#691](https://github.com/meshtastic/web/issues/691)

PR #1926, opened 2025-07-06
Update index.mdx

oliverkrystal:
Add information for GPS configuration on T-Deck Plus

## What did you change
Added Tx/Rx pin information

## Why did you change it
Trying to update T-Deck docks so that they have the information I was looking for instead of having to search forums.

### DIFF
--- a/docs/hardware/devices/lilygo/tdeck/index.mdx
+++ b/docs/hardware/devices/lilygo/tdeck/index.mdx
@@ -68,6 +68,11 @@ The following process will manually place the device into the Espressif Firmware
 
 With the device now in the Espressif Firmware Download mode, you can proceed with flashing using one of the supported flashing methods. It's generally recommended to use the [Web Flasher](https://flasher.meshtastic.org/). You can select "T Deck" from the device drop-down.
 
+## Configuring GPS on T-Deck Plus
+Referencing the [unit test code sample](https://github.com/Xinyuan-LilyGO/T-Deck/blob/master/examples/UnitTest/utilities.h) from LilyGo, you will need to set in Configuration -> Radio Config -> Position the GPIO pins for the GPS as follows:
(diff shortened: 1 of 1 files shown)


docs/hardware/devices/lilygo/tdeck/index.mdx
```diff
@@ -68,6 +68,11 @@ The following process will manually place the device into the Espressif Firmware
 
 With the device now in the Espressif Firmware Download mode, you can proceed with flashing using one of the supported flashing methods. It's generally recommended to use the [Web Flasher](https://flasher.meshtastic.org/). You can select "T Deck" from the device drop-down.
 
+## Configuring GPS on T-Deck Plus
+Referencing the [unit test code sample](https://github.com/Xinyuan-LilyGO/T-Deck/blob/master/examples/UnitTest/utilities.h) from LilyGo, you will need to set in Configuration -> Radio Config -> Position the GPIO pins for the GPS as follows:
```
pdxlocations, line 72:
```suggestion
You will need to set the GPS GPIO pins in Configuration -> Radio Config -> Position as follows:
```

Not sure we have to reference the external link.  If it works, it works?

oliverkrystal: I struggled to find the LilyGo repository for the unit while I was having power cycling issues.  Plus, more information and documentation is always better in my mind.  I'd like to leave it in.

pdxlocations:
I just looked at the source code and these pins are already set in the tdeck variant so there's no need to alter the configuration as you're describing.  Unless I'm missing something?
https://github.com/meshtastic/firmware/blob/195b7cc30a4337877e1e81a800820eca71d19a79/variants/t-deck/pins_arduino.h#L11-L12

PR #1933, opened 2025-07-14
Improve documentation around SPI radios and Docker compose usage

ywaf:
## What did you change
Improve documentation around SPI radios and Docker compose usage

## Why did you change it

Spent hours trying to find why the container wouldnt run the webui. I was using an alpine container. Its not documented anywhere except the Dockerfile that the alpine container lacks the webui. 

### Before
N/A

### After
N/A


rcarteraz:
@vidplace7 @danditomaso @jp-bennett Can you all review and determine if this is something that is needed still? This is quite old so not sure if we need to get it in or close it.

PR #1957, opened 2025-07-29
Update index.mdx

Albatross100541:


rcarteraz:
Not sure why this change would be required. I can only assume this was a mistake.

PR #1962, opened 2025-08-03
Update local-groups.mdx

esackbauer:
Added Austrian user group.


rcarteraz:
Hello -- Unfortunately, we will not be able to merge this request as is. The Meshtastic trademark guidelines do not allow for the trademark to be used as the domain name. See here: https://meshtastic.org/docs/legal/licensing-and-trademark/#domain-names

PR #1970, opened 2025-08-07
Update tips to describe sharing location to a secondary channel

RCGV1:
Replaced the "Creating a Private Primary with Default Secondary" tip with "Sharing Location on a Private Secondary Channel".

This instructs users how to share locations on a secondary channel, as this was not documented at all.

### DIFF
--- a/docs/configuration/tips.mdx
+++ b/docs/configuration/tips.mdx
@@ -58,6 +58,7 @@ The `position_precision` setting allows control of the level of precision for lo
 For detailed information on position precision settings and how to configure them, please refer to the [Position Precision documentation](/docs/configuration/radio/channels/#position-precision).
 
 ### Sharing Location on a Private Secondary Channel
+> This is a newer feature that only works for firmware 2.6.10+
 
 To share your location on a private secondary channel while keeping the default primary channel unencrypted, follow these steps:
 
@@ -83,7 +84,6 @@ Imagine you’re in the following channels:
 - Friends (Channel 2) can request your location manually, but it won’t receive automatic updates.
 - TutuTuesdayGroup (Channel 3) allows chatting but does not share or receive location data.
 
-> Note: You can be in up to 8 channels, but automatic location sharing is limited to one channel at a time. Other channels can access your location only through manual requests.
 ## Rebroadcast "Public" Traffic
 
 Meshtastic nodes will rebroadcast all packets if they share [LoRa modem settings](/docs/configuration/radio/lora#lora-config-values), regardless of encryption (unless [Rebroadcast mode](/docs/configuration/radio/device#rebroadcast-mode) is set to `LOCAL_ONLY`).
--- a/docs/configuration/tips.mdx
+++ b/docs/configuration/tips.mdx
@@ -78,7 +78,7 @@ Imagine you’re in the following channels:
 - Channel 2: “Friends” (location ON)
 - Channel 3: “TutuTuesdayGroup” (location OFF)
 
-### How It Works
+##### How It Works
 - CampChat (Channel 1) receives your live, automatic location updates because it’s the lowest-indexed channel with location sharing enabled.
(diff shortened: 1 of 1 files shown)


RCGV1:
Heavy inspo from feebs burning mesh docs
https://docs.burningmesh.org/en/guides/camp-channels-group-chats

docs/configuration/tips.mdx
```diff
@@ -57,25 +57,33 @@ The `position_precision` setting allows control of the level of precision for lo
...
+1. Disable location sharing on the primary channel (Channel 0) to prevent broadcasting your position to all users.
+2. Select a secondary channel where you want to share your location. Enable position sharing and choose your desired precision level.
+3. Understand automatic location updates:
+   - Only one channel receives your automatic, live location updates.
+   - This is the lowest-indexed secondary channel (excluding Channel 0) with location sharing enabled.
+   - If multiple secondary channels have location sharing enabled, only the one with the lowest index will receive automatic position broadcasts.
+   - Channels with location sharing enabled but not receiving automatic broadcasts can still be used for manual position requests.
 
-### Default Primary Frequency Slots by Region
+## Example Scenario
```
tekstrand: This needs to be a level 4 heading
```suggestion
#### Example Scenario
```

docs/configuration/tips.mdx
```diff
@@ -57,25 +57,33 @@ The `position_precision` setting allows control of the level of precision for lo
...
+Imagine you’re in the following channels:
 
-To quickly test this configuration, find and scan your region's QR[^1] from [this repository](https://github.com/meshtastic/meshtastic/tree/master/static/img/configuration/qr-private-primary-example/).  The example QR code will create a private primary channel and default Meshtastic secondary channel as demonstrated above. The private channel name may be updated if desired, but a new PSK should be generated before sharing with trusted nodes.
+- Channel 0: “Everyone” (location OFF)
+- Channel 1: “CampChat” (location ON)
+- Channel 2: “Friends” (location ON)
+- Channel 3: “TutuTuesdayGroup” (location OFF)
 
-[^1]: <QRCode />
+### How It Works
```
tekstrand: Should then be a level 5 or just bold text
```suggestion
##### How It Works
```

docs/configuration/tips.mdx
```diff
@@ -57,25 +57,33 @@ The `position_precision` setting allows control of the level of precision for lo
...
+- CampChat (Channel 1) receives your live, automatic location updates because it’s the lowest-indexed channel with location sharing enabled.
+- Friends (Channel 2) can request your location manually, but it won’t receive automatic updates.
+- TutuTuesdayGroup (Channel 3) allows chatting but does not share or receive location data.
 
+> Note: You can be in up to 8 channels, but automatic location sharing is limited to one channel at a time. Other channels can access your location only through manual requests.
```
tekstrand: This feels duplicative, I would suggest we remove it.

jschrempp:
Is this known to be true? I've been told before that a response to a location request would go out on Channel 0. 

If this PR is correct, then this is good news.

jschrempp (changes requested):
"To share your location on a private secondary channel while keeping the default primary channel unencrypted, follow these steps:"

Do you mean truly unencrypted, which is not standard, or do you mean "lightly encrypted with the default key".   If the primary channel has to be really unencrypted then this documentation change is about a slightly different situation than the currently described scenario.

skyycode-0:
I think it should be "lightly encrypted" and the such, as even the default primary channel is encrypted with a basic key.

rcarteraz:
> Is this known to be true? I've been told before that a response to a location request would go out on Channel 0.
> 
> If this PR is correct, then this is good news.

Previously, that was the case. But since https://github.com/meshtastic/firmware/pull/6920 you can.

jschrempp:
Maybe the new text should say, "as of release x.x.x " so that it is clear that this is new behavior.

tekstrand (approved):

PR #1984, opened 2025-08-17
Update that-one-time-at-defcon.mdx

jp-bennett:
Add addendum to blog post, pointing out that the attack was not entirely novel.


pdxlocations (approved):

rcarteraz:
Looks good. I would suggest adding a timestamp of when the update was added. So if merged today (as is the goal), then this under the addendum heading. 


_Updated 8/18/2025_

PR #1986, opened 2025-08-18
docs: fix typo that incorrectly names an already used pin for mosfet gate

lukasmalkmus:
https://meshtastic.org/de/docs/hardware/devices/heltec-automation/lora32/peripherals/#instructions-through-hole-mosfet states:

```
1. Solder a wire from the TXD pin on the GPS module to GPIO 48 on Heltec board. [...]
...
6. Solder a wire from Middle leg (lead 2, Gate) of NPN MOSFET to GPIO 48 of Heltec board. [...]
....
10. Set PIN_GPS_EN to 46 [...]
```

Since GPIO 48 is already soldered to GPS TXD and given the instruction to set PIN_GPS_EN to 46, I'm pretty sure bullet point six should actually say to solder to GPIO 46.


pdxlocations (approved):

pdxlocations:
The image and description agree with this fix

PR #1992, opened 2025-08-19
Update power.mdx

adrianTNT:
added default adc multipliers value for seeed xiao nrf52840

### DIFF
--- a/docs/configuration/radio/power.mdx
+++ b/docs/configuration/radio/power.mdx
@@ -112,6 +112,7 @@ Should be set to floating point value between 2 and 6
       | rak4631 | 1.73 |
       | rpipico | 3.1 |
       | rpipicow | 3.1 |
+      | seeed xiao nrf52840 | 3.1 |
       | station-g1 | 6.45 |
       | station-g2 | 4 |
       | tlora_v2_1_16 | 2 |


pdxlocations:
Thanks @adrianTNT, would you sign the the CLA so this can be merged?

adrianTNT:
> Thanks @adrianTNT, would you sign the the CLA so this can be merged?

I think I did now, did it work? I am new :)

pdxlocations:
Is this the variant you're adding?  I may be looking in the wrong place but I'm seeing 3.0 here.
https://github.com/meshtastic/firmware/blob/3b6eefa8bbfcb2ae02e4e311b190cc2dfef9a05b/variants/nrf52840/seeed_xiao_nrf52840_kit/variant.h#L166

adrianTNT:
@pdxlocations I don't think that is related. What I edited was adding `| seeed xiao nrf52840 | 3.1 |` inside file `power.mdx`:

pdxlocations:
I'm attempting to check your submission against the codebase.  Where does the number "3.1" come from?

adrianTNT:
@pdxlocations the `3.1` is the value I found correct when tunning my Xiao NRF's "ADC multiplier" in meshtastic power menu.
I added it to that file in order for this value to show up on Meshtastic website documentation.
Are you referring to making this value available as default for XIao NRF in the meshtastic source code? I don't know about that part.

rcarteraz:
> @pdxlocations the `3.1` is the value I found correct when tunning my Xiao NRF's "ADC multiplier" in meshtastic power menu. I added it to that file in order for this value to show up on Meshtastic website documentation. Are you referring to making this value available as default for XIao NRF in the meshtastic source code? I don't know about that part.

The default values in the docs are based on the codebase, so they should match. It would need to be changed in the code first, before we update it here.

adrianTNT:
@rcarteraz I was under the impression that Meshtastic website shows us what ADC multiplier values we can enter ourselves for various devices in case the Meshtastic firmware didn't properly show battery voltage. So in that case I wanted to add these instructions (`3.1 for XIao NRF`). If the right value already exists in meshtastic code, then there is no need for these custom values / instructions on Meshtastic website. Or did I get it all wrong?

rcarteraz:
> @rcarteraz I was under the impression that Meshtastic website shows us what ADC multiplier values we can enter ourselves for various devices in case the Meshtastic firmware didn't properly show battery voltage. So in that case I wanted to add these instructions (`3.1 for XIao NRF`). If the right value already exists in meshtastic code, then there is no need for these custom values / instructions on Meshtastic website. Or did I get it all wrong?

The values in the table are the default ADC from the variant for that device. As part of the calculation, you would enter that value into the “current adc value” if you’re using a device running the default. It will then calculate the override value to input into your device to help fix the battery percentage if not correct. This can vary from device to device and battery to batter. So, for this table the only value want there is whatever is in the variant for the device. So before we’d change it here, it would need to be changed there. Hopefully that helps clear it up. Thanks!

PR #2002, opened 2025-08-25
Add note about mqtt special chars when using uplink.

derfloh:
## What did you change
Add a note about using problematic chars in channel names when using MQTT.

## Why did you change it
I had a hard time figuring out why DMs and messages to LongFast where published to my MQTT server, but not messages to "Channel#1" - After spending a couple of hours on various things like factory reset, reflashing, messing up my config even further etc. I finally realized that my messages couldn't be published because my Channel Name was incompatible with MQTT. I hope that adding this note will prevent others going down the same rabbit hole as I did.


pdxlocations:
Sorry for the late review.  I think since a search for "mqtt # *" answers the question about why they might not work in channel names, I will close this PR, as it's not related to Meshtastic but the nature of MQTT.  Sorry for your frustration and I do appreciate every PR for the docs!  I hope you can understand.

Issue #2003, opened 2025-08-30
[Documentation Change Request]:  Module API - Getting Started page needs to mention want_response flag.


JohannSuarez:
### Request Details

allocReply() is only invoked if the received packet has want_response flag set to true.
Text messages sent from phone apps do not set this flag, and so the custom module resulting from following the steps doesn't reply.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

garthvh:
For a text message you want an ack back not another text message.

JohannSuarez:
Thanks for the quick reply! You’re right that ack confirms delivery, but they don’t invoke the template module's code. The ReplyModule (the starter template in the [docs](https://meshtastic.org/docs/development/device/module-api/)) is only called when the incoming packet has want_response = true. That’s the flag that triggers MeshModule::sendResponse() → allocReply(). 

Phone chat UIs set want_ack but not want_response, so the template suggested in the Module API page appears inactive unless you use a client that can set wantResponse=True. I made a PR that adds that information here: [https://github.com/meshtastic/meshtastic/pull/2004](https://github.com/meshtastic/meshtastic/pull/2004)

But pardon me if I'm missing important details here, I'm new.

EDIT: I was linking to my local copy of the docs lol

PR #2004, opened 2025-08-30
module-api.mdx - Important information on receiving response from a custom module

JohannSuarez:
## What did you change
The Module API's page needs additional information on the behaviour of allocReply(), which is overriden in ReplyModule.

## Why did you change it
The section could use an explanation on how allocReply is invoked in the background, and it provides clues on what to check if someone's custom module from this tutorial isn't replying back.

[Issue Link](https://github.com/meshtastic/meshtastic/issues/2003)

### Before

image

### After

image


pdxlocations (approved):
if GUVWAV says it's good, then it's good!

PR #2008, opened 2025-09-01
Added Saudi Arabia in region-by-country.mdx

a-alomairah:
information provided "AS IS" according to LoRaWAN Regional Parameters document

## What did you change
Added the frequencies used for LoRa in Saudi Arabia.

## Why did you change it
Information wasn't available before


fifieldt:
Hi @a-alomairah , thanks for adding this. We've previously found the lora alliance document to contain inaccuracies, so these days we prefer to get information from original sources from the country itself. Do you have the skills and time to find the relevant document from the Saudi authorities?  (non-English is fine!)

fifieldt:
It might be the "Communications Space and Technology Commission" https://www.cst.gov.sa/

a-alomairah:
Hi @fifieldt , Thanks for your reply!

I actually looked into the Saudi Arabia sources. And only could find the frequencies that match the EU_433 , it was the only one that had the Amateur category in it. (Although I think you need an Amateur license to operate in it)

I tried to look at this source but couldn't get a definitive answer. (probably my lack of understanding of the categories and which category does the use of Meshtastic fall into). So, I assumed the lora alliance did their research to officially verify the frequencies listed in their document.

This is the **[National Frequency Plan](https://spectrum-individual.cst.gov.sa/nfp/spectrumplan)** that i was talking about. 
Perhaps it is best for me to contact them to get a clearer answer.

fifieldt:
Thanks @a-alomairah , yes, that was our original assumption too. However we then started seeing cases where the document included extra bands that didn't exist, missed some that did, or the frequencies were wrong :( So, now we do extra checks before propagating the information.

The [National Frequency Plan](https://spectrum-individual.cst.gov.sa/nfp/spectrumplan) sounds like the right place ... but for some reason it's not loading for me :( If you have the time to look into it that would be amazing! If I get time, I'll try and help too.

PR #2010, opened 2025-09-03
Added Philippine regulations in region-by-country.mdx

skyycode-0:
Updated the docs with Philippines as a region and its respective frequencies.


rcarteraz:
This first needs associated pulls in both protobufs and firmware.

skyycode-0:
> This first needs associated pulls in both protobufs and firmware. 

If I may, what do you mean associated pulls in both? Since (from checking past commits) there seems to be mentions and the inclusion of the frequency in both protobufs and firmware.

skyycode-0:
@rcarteraz I've added the last edits and I think this can be merged now. :)

PR #2017, opened 2025-09-09
Add tip for users to improve the page

danditomaso:
Added a tip section encouraging users to edit the page directly.

## What did you change

Added tip box to top of page to tell users they can make changes using the `edit this page` link at the bottom of the page 

## Why did you change it

The layout of the list requires users scroll to the bottom in order to make changes, which may not be understood by people simply wanting to add new groups to the page.


pdxlocations:
in the opening paragraph, we say "If you're a group organizer with an online presence and wish to be included in this list, please edit this page directly or reach out to us on Discord to add your group."

Could we just add "(click the link at the bottom of the page)" or something similar there?

pdxlocations:
Was this closed intentionally?

rcarteraz:
> Was this closed intentionally?

Yes, the information is already on the page. An admonition is unnecessary.

pdxlocations:
> An admonition is unnecessary.

100% agree with this.  How about something like https://github.com/meshtastic/meshtastic/pull/2019

PR #2019, opened 2025-09-10
Add Link to Edit in Opening Paragraph

pdxlocations:


pdxlocations:
> I wonder if it's possible to link to the button itself, so it scrolls down. I feel like that would reinforce people using it lol. But this is acceptable too.

I like this idea.  We can add an id tag to the end of the page and jump right to it.
https://meshtastic-i0iko7g8j-meshtastic.vercel.app/docs/community/local-groups/#edit-this-page

rcarteraz:
> > I wonder if it's possible to link to the button itself, so it scrolls down. I feel like that would reinforce people using it lol. But this is acceptable too.
> 
> I like this idea. We can add an id tag to the end of the page and jump right to it. https://meshtastic-i0iko7g8j-meshtastic.vercel.app/docs/community/local-groups/#edit-this-page

Fantastic!

PR #2020, opened 2025-09-10
Update region-by-country.mdx

RandomLinoge:
## What did you change
Added Israel w/ LoRa Region and Regulatory Document. 
Fixed line spaces in several places.

## Why did you change it
Israel LoRa Region was missing from the table list.

### Before
image

### After
image


garthvh:
This is incorrect, Israel has a 1% duty cycle, the project can't endorse using the US region

PR #2014, opened 2025-09-08
Add docs for new CLIENT_BASE role

compumike:
**Please do not merge until the Firmware PR is shipped!**

Issue: https://github.com/meshtastic/firmware/issues/7863 "CLIENT_BASE mode: ROUTER for favorites, CLIENT otherwise (for attic/roof nodes!)"

Firmware PR: https://github.com/meshtastic/firmware/pull/7873

### `/docs/configuration/radio/device/#roles`:

image

image

image

-----

### `/docs/configuration/tips/`:

image

-----

### `/blog/choosing-the-right-device-role/`:

image

PR #2030, opened 2025-09-18
Add blog post: Demystifying ROUTER_LATE

erayd:
## What did you change

Added a new blog post describing the ROUTER_LATE role in detail, and how it fits with other roles.

## Why did you change it
Because I find I am constantly fielding questions about this role and clearing up misunderstandings, so an 'official' resource people can peruse to clear up those questions would be helpful.

### DIFF
--- a/blog/2025/September/demystifying-router-late.mdx
+++ b/blog/2025/September/demystifying-router-late.mdx
@@ -20,6 +20,8 @@ needs it's important to select the right role to ensure that the overall mesh
 is robust, does not get overloaded, and ensures the reliable transit of packets
 to where they need to go.
 
+[1]: /blog/choosing-the-right-device-role/
+
 {/* truncate */}
 
 ## What is the purpose of ROUTER_LATE?
@@ -234,5 +236,4 @@ No. Please don't do this - ROUTER_LATE is not intended as a mobile role, and
 using it in that manner will usually cause more problems than it solves.
 
 
-[1]: /blog/choosing-the-right-device-role/
 [2]: /blog/why-meshtastic-uses-managed-flood-routing/


jschrempp:
I read this about ROUTER_LATE, "Within that late window, it will still use the same timing behaviour as CLIENT."

I don't understand the "same timing behaviour as CLIENT". I thought that ROUTER_LATE would always rebroadcast a packet in the late window, this leaves me wondering. Maybe add a link to a place that describes the timing behaviour. Do you mean to say that even in the late window the ROUTER_LATE will choose a broadcast time based on the received SNR?

jschrempp:
"ROUTER_LATE is designed as an infrastructure role for serving parts of larger or more complex meshes that do not have visibility to existing ROUTER sites." 

To me, this statement implies that all nodes in "larger or more complex" meshes should have visibility to a ROUTER node. Thus, if they don't, then maybe a ROUTER_LATE (or ROUTER) node should be put in place. I don't think that is your intention.

jschrempp:
In the table you say this "and can cause non-infrastructure roles to cancel their own rebroadcast." in the description of the early window but not the other windows. Later you write that this is true of all windows. I would remove this sentence from the description of the early window or add it to the other two window descriptions. By having it on only one description it implies there is some difference.

jschrempp:
"If you insist on using this role as a roof node anyway, "

I would suggest that another mitigation is to reduce the Transmit Power of such a node to a very low level, just enough to cover the intended service area, perhaps 3 or 4. This will limit the impact of this node on the surrounding mesh.

jschrempp:
It would be good to add a few words about why the REPEATER role should not be used. To a noob this also sounds like a good idea for a roof top node.

jschrempp:
"No. In this case, please deploy a CLIENT node nearby."  maybe add "and increase your hop limit by 1"?

rcarteraz:
Drafting until a release is out with client_base due to mention and for #2014 to be merged so we have some documentation on that as well.

erayd:
@jschrempp Thanks for your feedback 🙂

> I thought that ROUTER_LATE would always rebroadcast a packet in the late window, this leaves me wondering.

It doesn't rebroadcast in the late window at all, unless it hears another note rebroadcasting. This is one of the most common misconceptions about the role.

> Maybe add a link to a place that describes the timing behaviour.

There is nowhere authoritative that describes the timing behavior properly. The whole point of *this* blog post is to be such an authoritative source.

> Do you mean to say that even in the late window the ROUTER_LATE will choose a broadcast time based on the received SNR?

*Every* meshtastic role uses SNR-biased timing, always. This includes ROUTER_LATE, both within the default window and within the late window. 

> To me, this statement implies that all nodes in "larger or more complex" meshes should have visibility to a ROUTER node. Thus, if they don't, then maybe a ROUTER_LATE (or ROUTER) node should be put in place. I don't think that is your intention.

That is precisely my intention. Large / complex meshes without mandatory-rebroadcast nodes tend to have parts of the mesh miss messages, due to the rebroadcast-cancelling behaviour of clients. Adding infrastructure nodes fixes this. The whole reason that ROUTER_LATE exists in the first place is because the default CLIENT behaviour is incapable of acting as a reliable relay.

> By having it on only one description it implies there is some difference.

The distinction is deliberate, because there *is* a difference. A packet in the early rebroadcast window preempts *all* other non-infrastructure nodes. A packet anywhere else doesn't preempt, and allows the normal SNR bias + random timing to dictate which nodes will rebroadcast a packet.

> I would suggest that another mitigation is to reduce the Transmit Power of such a node to a very low level, just enough to cover the intended service area, perhaps 3 or 4. This will limit the impact of this node on the surrounding mesh.

I deliberately did not mention this, because in my experience even extremely low levels of power can go quite a long way - e.g. a 1dBm signal from the roof of my house  an still be received halfway across the city I live in. I do not wish to give the impression that dropping the power is a reliably means of mitigating interference, because it isn't (although it does help), and my preference is to discourage use of ROUTER_LATE as a rooftop node entirely. That is what the new CLIENT_BASE role is designed for.

> It would be good to add a few words about why the REPEATER role should not be used. To a noob this also sounds like a good idea for a roof top node.

What did you have in mind?

> Maybe add "and increase your hop limit by 1"?

Why? Increasing hop limit is not necessary unless the default is insufficient to reach the node with which you are wanting to communicate. In most cases, the default is sufficient.

jschrempp:
@erayd Thanks for the thoughtful comments on mine. 

>It doesn't rebroadcast in the late window at all, unless it hears another node rebroadcasting. This is one of the most common misconceptions about the role.

Oh! News to me. Thanks.

>A packet anywhere else doesn't preempt, and allows the normal SNR bias + random timing to dictate which nodes will rebroadcast a packet.

I was perhaps hung up on the distinction between "preempt" and the "normal" process where a node decides not to rebroadcast. In my mind when a node in the default window decides not to rebroadcast because it heard the packet from someone else, it has been "preempted". I'll have to think of another word to use for that situation. Perhaps "suppressed".

>>It would be good to add a few words about why the REPEATER role should not be used. To a noob this also sounds like a good idea for a roof top node.

>What did you have in mind?

Unfortunately I am still learning all this. In my experience the asymmetry in nodes causes new people to receive from powerful nodes but be unable to send into the mesh. They look for solutions and REPEATER or ROUTER sounds like "oh, I need that". You do a good job of beating down the idea of ROUTER and ROUTER_LATE. And CLIENT_BASE is probably what they need. I just thought the same kind of commentary on REPEATER would be helpful.

Again, thanks for a very informative blog post.

jschrempp:
@erayd 

These words from the official docs tell a slightly different story of ROUTER_LATE. Are they inaccurate? If so, I can work to change them.

https://meshtastic.org/docs/configuration/radio/device/#:~:text=ROUTER_LATE,t%20benefit%20the%20broader%20mesh.
"Infrastructure node that always rebroadcasts packets once but only after all other modes, ensuring additional coverage for local clusters. Visible in Nodes list."

From what you say ROUTER_LATE might actually transmit in the Default Window if it happens to win the rebroadcast race.

erayd:
> @erayd
> 
> These words from the official docs tell a slightly different story of ROUTER_LATE. Are they inaccurate? If so, I can work to change them.
> 
> https://meshtastic.org/docs/configuration/radio/device/#:~:text=ROUTER_LATE,t%20benefit%20the%20broader%20mesh. "Infrastructure node that always rebroadcasts packets once but only after all other modes, ensuring additional coverage for local clusters. Visible in Nodes list."

The part of the docs you are quoting is an oversimplification. I'd leave it the way it is, because it's easier for users to reason about. It's not technically quite right, but it's close enough for users who are trying to determine which role makes sense for them to be using, and understand how it behaves.

> From what you say ROUTER_LATE might actually transmit in the Default Window if it happens to win the rebroadcast race.

Yes, that is correct.

erayd:
> Unfortunately I am still learning all this. In my experience the asymmetry in nodes causes new people to receive from powerful nodes but be unable to send into the mesh. They look for solutions and REPEATER or ROUTER sounds like "oh, I need that". You do a good job of beating down the idea of ROUTER and ROUTER_LATE. And CLIENT_BASE is probably what they need. I just thought the same kind of commentary on REPEATER would be helpful.

I've added a comment about ROUTER & REPEATER in the FAQ section regarding rooftop nodes. Does this address what you were wanting addressed?

jschrempp:
> I've added a comment about ROUTER & REPEATER in the FAQ section regarding rooftop nodes. Does this address what you were wanting addressed?

Looks good.

GUVWAF (approved):
Very nice and elaborate blog post!

blog/2025/September/demystifying-router-late.mdx
```diff
@@ -0,0 +1,238 @@
+---
+title: "Demystifying ROUTER_LATE"
+description: "What is the purpose of the ROUTER_LATE role, and when should it be used?"
+slug: demystifying-router-late
+authors: erayd
+tags: ["Meshtastic", "devices", "roles", "routing", "infrastructure"]
+date: 2025-09-18T16:42+1200
```
rcarteraz: I think the +1200 is causing issues, you can drop that. Also, let's update to today's date.

erayd: Done - it's now UTC.

erayd:
Looks like there is a bug in whatever software is rendering the "above the fold" preview for this - it's failing to correctly render the link in the first paragraph. 

When opening the full post, the link is rendered correctly.

Is this something that will be fixed in the rendering software, or should I update the mdx file to use a different linking syntax? If so, what syntax is actually supported? I'm currently using standard markdown references.

Edit: Have moved the reference above the `{*truncate*}` tag to see if this helps. (Further edit: it did.)

PR #2031, opened 2025-09-19
Update device.mdx Changed description of ROUTER_LATE

jschrempp:
Note that ROUTER_LATE first acts as a CLIENT.

## What did you change
ROUTER_LATE incorrectly said that it would rebroadcast after all other nodes. 

## Why did you change it
ROUTER_LATE first acts as a normal Client in the default rebroadcast window. It only rebroadcasts in the Late window if it heard another node broadcast first in the default window.

This is a significant behavior. It is not documented anywhere else.


docs/configuration/radio/device.mdx
```diff
@@ -27,7 +27,7 @@ The device config options are: Role, Rebroadcast Mode, GPIO for User Button, GPI
...
 | ROUTER         | Infrastructure node for extending network coverage by always rebroadcasting packets once. Visible in Nodes list.                                                 | Best positioned in strategic locations to maximize the network's overall coverage. Device is shown in topology.                                             |
-| ROUTER_LATE    | Infrastructure node that always rebroadcasts packets once but only after all other modes, ensuring additional coverage for local clusters. Visible in Nodes list. | Ideal for covering dead spots or ensuring reliability for a cluster of nodes where placement doesn’t benefit the broader mesh. Device is shown in topology. |
+| ROUTER_LATE    | Infrastructure node that always rebroadcasts packets once. First it acts as a normal client. If any other node rebroadcasts the message then it will still rebroadcast later than any other node would, ensuring additional coverage for local clusters. Visible in Nodes list. | Ideal for covering dead spots or ensuring reliability for a cluster of nodes where placement doesn’t benefit the broader mesh. Device is shown in topology. |
```
rcarteraz: Personally, I don't think this helps make it any clearer or easier for the user to understand. In fact, I think @erayd said it best that this is an oversimplification but probably easier for everyone to understand. 

https://github.com/meshtastic/meshtastic/pull/2030#issuecomment-3309989022

> The part of the docs you are quoting is an oversimplification. I'd leave it the way it is, because it's easier for users to reason about. It's not technically quite right, but it's close enough for users who are trying to determine which role makes sense for them to be using, and understand how it behaves.

jschrempp: I could use some help on this. I understand your comment about this being what some people would consider a nit, and not needed for the general public new to Meshtastic. OTOH, this is a significant insight into how ROUTER_LATE actually works and I think it's important information for more advanced users who want to really understand the roles without having to read the code. I've told a few people about this behavior and they were interested; I think it is valuable for some people.

So, where should this behavior be documented? I searched the documentation and this is the only page that mentions ROUTER_LATE, so I put it here. If you have an idea of where this information would fit better in the docs, I'm willing to add it there, or start a new page if that's the way to go. 

Or if you think this kind of information doesn't belong in the official public docs, let me know that too.

Thanks.

erayd: > So, where should this behavior be documented? I searched the documentation and this is the only page that mentions ROUTER_LATE, so I put it here.

That behavior is now fully documented [here](https://meshtastic.org/blog/demystifying-router-late/). The point of that blog post is to ensure that users can fully understand the role, in much more depth than the single sentence we have space for in the roles summary table.

> I understand your comment about this being what some people would consider a nit, and not needed for the general public new to Meshtastic. OTOH, this is a significant insight into how ROUTER_LATE actually works and I think it's important information for more advanced users who want to really understand the roles without having to read the code. I've told a few people about this behavior and they were interested; I think it is valuable for some people.

IMO, the summary table should provide a bite-sized snippet that is extremely easy for people to reason about, and will lead to them making (mostly) correct choices about which role to use even if they do not read anything else. Hence why I feel that the oversimplification is appropriate there. If we get into the weeds on timing & mutable behaviour in that table, then users end up confused, because they do not - without further research - understand how that timing and behaviour actually works. It's one of those situations where a technically incorrect statement actually conveys the *spirit* of what the ROUTER_LATE role does better than an accurate, more detailed summary would.

It could potentially be useful to link the ROUTER_LATE blog post from below that table (next to the "choosing the right device role" link), or alternatively from within the choosing-a-role post itself. That would make the in-depth information more discoverable for those who wish to dive down the rabbit hole.

jschrempp: The link to the blog post is a great idea. I did that.

PR #2036, opened 2025-09-21
Update local-groups.mdx

Xaositek:
## What did you change
Add LZMesh in SW Missouri to Local Group


Xaositek:
@rcarteraz Whenever you have a moment, this one has no dependencies.

PR #2044, opened 2025-09-27
Add TinyTronics purchase links

codesoap:
## What did you change
I have added TinyTronics as a source for the Heltec v3 and mesh node T114. I have verified that the T114 can be bought and works.

## Why did you change it
I was missing a vendor in europe, which ships a bit faster than AliExpress.


rcarteraz:
We are only adding official links provided by the vendors at this time. Thank you though.

PR #2049, opened 2025-10-05
Update ota.mdx

jschrempp:
Make the change to Packets Receipt Notification a normal part of the process. 

## What did you change
Moved the recommendation to change packets receipt notification to a recommended step rather than an after failure action since a failure can brick the device.

## Why did you change it
I did this because a failure during OTA update can fail in a way that requires physical access to the device to recover. It seems better to make this a regular step rather than expose the user to a greater chance of failure.

### DIFF
--- a/docs/getting-started/flashing-firmware/nrf52/ota.mdx
+++ b/docs/getting-started/flashing-firmware/nrf52/ota.mdx
@@ -66,22 +66,20 @@ OTA firmware updates are available for Android using an older release of the mor
 
 #### Apple
 
-OTA firmware updates are available on iOS & iPadOS using the nRF Device Firmware Update App available through the [Apple App Store](https://apps.apple.com/us/app/nrf-device-firmware-update/id1624454660)
+OTA firmware updates are available on iOS & iPadOS using the nRF Device Firmware Update App available through the [Apple App Store](https://apps.apple.com/us/app/nrf-device-firmware-update/id1624454660). Note that if the OTA update fails your device may stop working until you gain physical access to it and upload firmware over USB.
 
 1. Download the firmware release you wish to install from the [Meshtastic Download Page](/downloads), [Meshtastic GitHub](https://github.com/meshtastic/firmware/releases), or via the iOS or iPadOS app.
 2. Unzip the firmware folder
 3. Open the nRF DFU App and select the correct device firmware file (will end with -ota.zip)
-4. In the nRF DFU Settings (in the upper right hand corner) enable "Packets receipt notification" and set the "Numer of Packets" to 10 or less. If your device is hard to get to, then start with a low number until you know it works; if the number is too high you might lock up your device and require a physical connection to flash the firmware.
-5. Connect to your device
-6. Upload the firmware. This can take 5 minutes.
+4. Connect to your device
+5. Upload the firmware
 
 :::Info
-On the iPhone interrupting the upload can cause it to fail. You should consider turning on Airplane mode and disabling WiFi to prevent interruptions during the upload. Also, the iPhone's auto-lock feature could potentially interrupt the Bluetooth firmware upload. To avoid this, occasionally tap on your screen, or temporarily set the auto-lock to "Never" during the upload process to ensure that the phone stays awake and the upload completes without interruption. 
+With the iPhone, interrupting the OTA upload can cause it to fail and leave your device in a non-working state. If your node is difficult to reach physically (such as on a roof top) you should follow the steps below before beginning the firmware upload to increase your chances of a successful OTA update.
 :::
+1. In the nRF Device Firmware Update App application Settings (in the upper right hand corner) enable "**Packets receipt notification**" and set the "**Numer of Packets**" to 5. 
+2. Turn on Airplane mode and disable WiFi to prevent interruptions during the upload. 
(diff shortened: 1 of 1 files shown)


docs/getting-started/flashing-firmware/nrf52/ota.mdx
```diff
@@ -71,10 +71,11 @@ OTA firmware updates are available on iOS & iPadOS using the nRF Device Firmware
 1. Download the firmware release you wish to install from the [Meshtastic Download Page](/downloads), [Meshtastic GitHub](https://github.com/meshtastic/firmware/releases), or via the iOS or iPadOS app.
 2. Unzip the firmware folder
 3. Open the nRF DFU App and select the correct device firmware file (will end with -ota.zip)
-4. Connect to your device
-5. Upload the firmware
+4. In the nRF DFU Settings (in the upper right hand corner) enable "Packets receipt notification" and set the "Numer of Packets" to 10.
```
pdxlocations: Instead of using "10" as a rule, could we suggest that a lower number may be helpful?  I've had luck with the default, while others swear it has to be 5.

rcarteraz:
This update isn't really helpful in my opinion, it takes simple instructions and convolutes them to the point where most readers will stop reading. If you are wanting to convey that they may need to change the packet count then I would simply add a quick note for that.

jschrempp:
> This update isn't really helpful in my opinion, it takes simple instructions and convolutes them to the point where most readers will stop reading. If you are wanting to convey that they may need to change the packet count then I would simply add a quick note for that.

A note already exists that the user "may" want to lower the packet count. The problem I see is that someone following the instructions as they are will see the note as optional, use the app without the change, and brick their node. If their node is hard to reach, then the instructions have led them to a very bad state. It happened to me and I've heard others say it happened to them. 

I would be ok with changing the instructions to say this step is mandatory and then have a note that the user "may consider setting the packet count higher if they are willing to risk losing the node". I'd want to say what the advantage is to a higher packet count. Does that make sense to you?

rcarteraz:
Nothing about what you are proposing is mandatory as not everyone has these issues. I can update multiple devices without making any changes to the packet counts. Not to mention you've taken simple instructions and made them way more complicated and requiring a lot more reading. I'm just not sure I see any value in this update. If you feel like an update is needed to better help people understand they may need to perform these steps, I get that but it needs to be simple. The change you've added is way too verbose.

jschrempp:
Good points. I have changed my proposed text to leave the main flow for firmware upload as it was. I changed the section on "if the OTA fails" to be "if your node is hard to get to, then do these steps to increase your chance of success".

docs/getting-started/flashing-firmware/nrf52/ota.mdx
```diff
@@ -66,19 +66,20 @@ OTA firmware updates are available for Android using an older release of the mor
...
 1. Download the firmware release you wish to install from the [Meshtastic Download Page](/downloads), [Meshtastic GitHub](https://github.com/meshtastic/firmware/releases), or via the iOS or iPadOS app.
 2. Unzip the firmware folder
 3. Open the nRF DFU App and select the correct device firmware file (will end with -ota.zip)
 4. Connect to your device
 5. Upload the firmware
 
-The iPhone's auto-lock feature could potentially interrupt the Bluetooth firmware upload. To avoid this, occasionally tap on your screen, or temporarily set the auto-lock to "Never" during the upload process to ensure that the phone stays awake and the upload completes without interruption.
-
-If the update fails, you may find that adjusting the packet settings can help:
+:::Info
```
rcarteraz, line 77: The admonition isn't necessary

docs/getting-started/flashing-firmware/nrf52/ota.mdx
```diff
@@ -66,19 +66,20 @@ OTA firmware updates are available for Android using an older release of the mor
...
+With the iPhone, interrupting the OTA upload can cause it to fail and leave your device in a non-working state. If your node is difficult to reach physically (such as on a roof top) you should follow the steps below before beginning the firmware upload to increase your chances of a successful OTA update.
+:::
+1. In the nRF Device Firmware Update App application Settings (in the upper right hand corner) enable "**Packets receipt notification**" and set the "**Numer of Packets**" to 5.
```
rcarteraz, line 80: We've already referred the name of the app in full, followed by a followup of nRF DFU app, we can do the same here as well.

docs/getting-started/flashing-firmware/nrf52/ota.mdx
```diff
@@ -66,19 +66,20 @@ OTA firmware updates are available for Android using an older release of the mor
...
-
-If the update fails, you may find that adjusting the packet settings can help:
+:::Info
+With the iPhone, interrupting the OTA upload can cause it to fail and leave your device in a non-working state. If your node is difficult to reach physically (such as on a roof top) you should follow the steps below before beginning the firmware upload to increase your chances of a successful OTA update.
+:::
+1. In the nRF Device Firmware Update App application Settings (in the upper right hand corner) enable "**Packets receipt notification**" and set the "**Numer of Packets**" to 5. 
+2. Turn on Airplane mode and disable WiFi to prevent interruptions during the upload. 
+3. The iPhone's auto-lock feature could potentially interrupt the Bluetooth firmware upload. To avoid this, occasionally tap on your screen, or temporarily set the auto-lock to "Never" during the upload process to ensure that the phone stays awake and the upload completes without interruption.
```
rcarteraz, line 82: This is pretty duplicative of the opening paragraph. I suggest condensing it there or much more condensed reminder her.

PPPDUD (changes requested):
You left a typo by saying "numer of packets", perhaps you meant _number_ of packets?

PR #2058, opened 2025-10-09
Update local-groups.mdx

jeremyfuksa:
Added Kansas City network site.


jeremyfuksa:
I think that's fair. We can revisit when the site is no longer a personal site and has a larger community running it.

PR #2061, opened 2025-10-11
local communities: update germany/berlin

l5yth:
## What did you change

local communities: update germany/berlin

## Why did you change it

to reflect the berlin community resources


l5yth:
Done, thanks for checking!

PR #2063, opened 2025-10-12
Update range-test.mdx - Added that Unset is the default on Apple apps

jschrempp:
Added that "Unset" is the default in the Apple apps; 0 is not an option.

## Why did you change it
Documentation said 0 is the default Sender interval, but that is not a choice in the Apple apps.

### DIFF
--- a/docs/configuration/module/range-test.mdx
+++ b/docs/configuration/module/range-test.mdx
@@ -33,7 +33,7 @@ Enables the range test module. **Both Sender and Receiver must have the module e
 
 ### Sender Interval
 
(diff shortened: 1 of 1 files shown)


rcarteraz:
0 is default, default is unset. The apple app just makes it easier for people to understand that.

jschrempp:
My change was to the part of the documentation that describes how the Apple apps work. I am not suggesting a change to the app, I'm making a change to the documentation so that it accurately reflects what the user sees in the app. I think it is confusing to read that the default is 0, yet the app says "unset". My change to the documentation resolves that confusion.

PR #2064, opened 2025-10-12
Update local-groups.mdx

aidan-lemay:
Adding ROCMesh to the list

## What did you change
Added ROCMesh to the list of community links

Add ROCMesh to the list of community links

## Why did you change it

New group we are founding that we would like membership into


aidan-lemay:
Branding has been updated to be fully custom!

PR #2080, opened 2025-10-27
Update web-client.mdx - Adding SBC Linux updating process. Round 2

Will-wastelander:
This is related to https://github.com/meshtastic/web/pull/559/

## What did you change

Added info on how to update WebUI for MeshtasticD.


docs/software/web-client.mdx
```diff
@@ -49,3 +49,25 @@ The method with the least platform support, which uses the [Web Serial API](http
 The source code for the WebUI can be found on our [GitHub](https://github.com/meshtastic/web)
 
 Instructions for building and running the project can be found in the repo's readme.
+
+## SBC Linux manual update
```
rcarteraz: Maybe @vidplace7 can chime in here, but I imagine the process is probably the same for regular linux? Can we change the Heading to just Updating on Linux? Or something like that?

Will-wastelander (commented):
Removed SBC as this would apply to any Linux install.

PR #2102, opened 2025-11-14
Update telemetry.mdx

vjx:
fix a typo, air resistance -> air quality

https://www.bosch-sensortec.com/products/environmental-sensors/gas-sensors/bme680/


rcarteraz:
The sensor gives you raw readings for temperature, humidity, pressure, and gas resistance (in ohms). Air quality is computed.

PR #2104, opened 2025-11-15
Add warning for Lora v1 support

radiskis:
Added warning about v1 tlora support


rcarteraz (changes requested):
Can you add this to the existing warning? Also, rather than a date can you put the corresponding version number?

rcarteraz (changes requested):
Better to remove the V1 entirely since it's no longer supported.

radiskis:
I would suggest keeping historical information to aid people with older hardware.

PR #2105, opened 2025-11-16
Fix formatting in RAK Wireless devices table

Mainstreetjay:
New Row: Added RAK3112 with a query param ?rakbase=RAK3112to match the pattern (this assumes the linked page uses URL params for filtering, as in the originals).
•  Description: Kept it concise and Meshtastic-relevant, highlighting its LoRaWAN focus for mesh networking.


rcarteraz:
This seems to be just a link to a non-existent page. I'm going to close this for now though as we have a draft update for this, but we're waiting for the starter kit to be available for purchase. We'll be adding in the near future.

PR #2114, opened 2025-11-19
Add CroMesh in Croatia

marinfyi:
## What did you change
i added CroMesh into Local Groups, a croatian group about meshtastic

## Why did you change it
to make meshtastic users aware of CroMesh

### Before
image

### After
image


rcarteraz (changes requested):
Hi! Thanks for contributing. It appears the name of the group here is CroMesh, but on the website it's being called "Meshtastic Croatia", which would be violation of the trademark guidelines. Please review the link below:

marinfyi:
> Hi! Thanks for contributing. It appears the name of the group here is CroMesh, but on the website it's being called "Meshtastic Croatia", which would be violation of the trademark guidelines. Please review the link below:
> 
> https://meshtastic.org/docs/legal/licensing-and-trademark/#social-media

we removed the meshtastic branding, could you take a look again please?

PR #2115, opened 2025-11-20
fix: replace gitignore, configs for crowdin

jamesarich:
- uncomment json sources/translations
- add locale dropdown
- replace `/i18n` in gitignore


jamesarich:
Merging this to stop the unnecessary action runs.

PR #2119, opened 2025-11-21
fix: crowdin translations

danditomaso:
## What did you change
This PR fixes has a number of fixes included in it: 
- The locale specific translations from crowdin were downloaded in order to run the server in development mode. Its important these are added to source control to avoid having to query Crowdin each time the site is booted.
- 8 fixes were made to the random `.mdx` files in the `DE` locale. They were preventing the server from running in development mode and were due to improperly closed HTML tags within the source .mdx files. 
- Removed the i18n folder from gitignore file, as mentioned above we need to start storing the translation files next to the server source code in our github, and updating it through an out of band github action.

Allow the German site translations to work as expected. If a page isn't translated the site will fall back to using english.

## Why did you change it

It didn't work previously, I helped James with getting this fixed and over the line.

### Before
image

### After
image

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -169,7 +169,6 @@ const config = {
   future: {
     v4: {
       useCssCascadeLayers: false,
-
     },
   },
 };


fifieldt:
Q) How will this handle languages that are partially translated? Will badly/partially translated pages end up online/in search indexes?

i18n/de/docusaurus-plugin-content-docs/current/about/faq.mdx
```diff
@@ -0,0 +1,264 @@
...
+slug: /faq
+sidebar_position: 3
+description: "Finden Sie Antworten auf Ihre Meshtastic Fragen: Wie beginnen, Geräte Einstellungen bis hin zu erweiterten Funktionen. Erfahren Sie mehr über App-Kompatibilität, Kanalfreigabe und die Vorteile der Amateurfunk-Lizenz. Tauchen Sie ein in unsere häufig gestellten Fragen (FAQ)."
+---
+
+import { FaqAccordion } from "/src/components/FaqAccordion";
+import { FaqStructuredData } from "/src/components/FaqStructuredData";
+import QRCode from "@site/docs/blocks/_qr-code.mdx";
+
+export const Faq = {
```
fifieldt, line 13: The FAQ looks untranslated. Looks like it wasn't picked up by crowdin.

i18n/zh/docusaurus-plugin-content-docs/current/about/_category_.yml
```diff
@@ -0,0 +1,7 @@
+position: 1 #float position is supported
```
fifieldt, line 1: As a user of this language, it's frustrating to land on a page like this that makes it seem like a translation exists, but it actually doesn't. Better to not publish.

i18n/es/docusaurus-plugin-content-docs/current/getting-started/index.mdx
```diff
@@ -0,0 +1,136 @@
+---
```
fifieldt, line 1: There's only a single page in this language (ES). Doesn't seem justified to put the /es langcode online.

fifieldt:
> Q) How will this handle languages that are partially translated? Will badly/partially translated pages end up online/in search indexes?

To answer my question, it seems like the answer is: yes. If merged, this patch will create indexable pages with incomplete translations for all of the languages for which docusaurus is currently configured. This is a poor user experience that should be avoided.

Suggested fix is @ https://discord.com/channels/867578229534359593/871540428510474302/1440529541759111272

fifieldt (changes requested):
^ per comment

i18n/de/docusaurus-plugin-content-docs/current/about/faq.mdx
fifieldt: (this is for DE)

danditomaso:
After chatting with James this isn't the best solution for this issue. Closing PR

PR #2125, opened 2025-11-24
Add t-beam 1.1 case

samm-git:
## What did you change

I added information about 3d case for the t-beam 1.1+ boards. Only referenced case is for the old (v0.7) board, which uses a different layout and is not compatible with new boards, which i learned in the hard way after printing :) 

## Why did you change it

Because I don't want other users to waste filament :) Also, the version i found fits the device perfectly

### Before
image

### After
image


docs/community/enclosures/tbeam.mdx
```diff
@@ -2,13 +2,17 @@
...
 ---
 
+There are [different revisions](/docs/hardware/devices/community-supported/lilygo/tbeam/)
+of the LILYGO® TTGO T-Beam Devices. On v0.7 (no longer in production), the antenna
+is aligned on the center of the board, on v1.1+ - on the right side.
+
 ## Created by tropho/TonyG
 
-### T-Beam V5 Case
+### T-Beam V5 Case for T-Beam v0.7
```
rcarteraz, line 15: Not sure where you're getting the v0.7 from. 
Screenshot 2025-12-02 at 1 55 46 PM

samm-git: from printing and view. I printed it and found that it does not match my t-beam.

samm-git: You can find other comments in the same model. So learned it in the hard way :)

docs/community/enclosures/tbeam.mdx
```diff
@@ -24,6 +28,16 @@ Purchase Links:
...
 - (x4) M2x4mm screws (no nuts) to secure T-Beam to the frame
 
-<img src="/img/enclosures/3dp-tropho-tbeam.webp" width="400" align="left" />
+<img src="/img/enclosures/3dp-tropho-tbeam.webp" width="400" />
 
 <!-- ![TrophoTbeam](/img/enclosures/3dp-tropho-tbeam.webp) -->
+
+## T-Beam v1.1+ created by Nyptus
+
+Download from [Printables](https://www.printables.com/make/2964198)
```
pdxlocations: Could we link to the model page instead of a make?

```suggestion
Download from [Printables](https://www.printables.com/model/1457112-case-for-lilygo-t-beam-v12)
```

PR #2128, opened 2025-11-29
Update MicroSD card slot info for SenseCAP Indicator

kristianrl:
## What did you change
Update MicroSD card slot info for SenseCAP Indicator

## Why did you change it
https://github.com/meshtastic/device-ui/issues/233#issuecomment-3535067731

### Before
-

### After
-


rcarteraz (changes requested):
Way too verbose when simply "not currently supported" could be used. Work is being done to add support via companion firmware for the co-processor.

PR #2133, opened 2025-12-02
Update local-groups.mdx

javicisro:
Added Romania local Groups


pdxlocations (changes requested):
Thanks for the submission, however I cannot merge any group that does not comply with the trademark guidelines.  Meshtastic [Location] is not permitted.  See:

Additionally we are trying to only allow one link per community group.  I would suggest only adding the website, since the social pages are linked from there.

Please make the required changes and I look forward to adding your group to the community page.

PR #2145, opened 2025-12-09
Update local-groups.mdx

Talie5in:
## What did you change
Added Australia / South Australia

## Why did you change it
It was missing? :D


Talie5in:
> I don't believe this logo is allowed use of the trademark.
> 

Fixed

image

PR #2151, opened 2025-12-15
fix: added vertical tabs to linux page

danditomaso:
## What did you change

This adds vertical tabs ONLY to the `linux-install` component. This was added after being requested by @vidplace7, the  available apps list overflows the screen and requires horizontal scrolling leading to a poor user experience.  

## Why did you change it

I was asked to make this change.

### Before
image

### After

after


pdxlocations:
Using chrome with default font size, I still see the horizontal layout (with scrollbar) when the window is over 1280px wide.  Anything narrower and the vertical tabs appear.  Is this intended behavior?

pdxlocations:
> Is this intended behavior?

Read through the discord comments

PR #2160, opened 2025-12-22
Revise Meshtastic Harbor Scale Integration documentation

Adrelien:
Updated the Meshtastic integration documentation to reflect the new Harbor Lighthouse agent features and installation instructions.

## What did you change

So this need a bit careful review. Before the idea was the run a python app then interact with the meshtastic package to collect data every certain amount of time then send it to the cloud, this was making a lot of users give up early one, most of them are not coders so they had not much experience with git or even python. Also different python version could cause come compatibility issues, pip install might not work for some. Too many steps. So the idea now might raise some eyebrows but everything is open source and anyone can review everything. 
Lighthouse is a collection agent built in golang to Harbor Scale it comes with some native data collection but the feature we used is the exec collector, any script or code that prints an array of json data can be collected and sent to the cloud to be visualized 
https://github.com/harborscale/harbor-lighthouse
Github actions will compile the code into an executable, the command is getting downloading it from https://get.harborscale.com and install it. the code for the install command also is open source 
https://github.com/HarborScale/harbor-lighthouse-installer
It looks suspicious on the surface but it is seamless way to install. 

with that the same idea was done, the same python code was changed, removed the gui, added auto detection for comport, and printed all the results so lighthouse can take it and sent it to the cloud 
https://github.com/HarborScale/harbor-meshtastic
the main.py would be compiled into an executable and then used with lighthouse to be seamless data collection 
https://get.harborscale.com/meshtastic install that complied exe from GitHub 
All the code also for the installer 
https://github.com/HarborScale/harbor-lighthouse-installer

Yes, a lot might not like to have executable installed on their pc but everything is open source and anyone can verify it. if you feel like it is too risky to list on the official docs it is totally understandable, you can reject this PR. the old GUI legacy way still works

NOTE: This also could be improved to collected data from MQTT or TCP IP instead of comport, A similar driver or even same driver can be configured with param to be passed to tell which way to collect data. I just worked too hard to get this done. So it will be a while since I visit the topic again for MQTT or TCP IP support
## Why did you change it
Streamline a base station data collection to the cloud to visualize the network status.

### DIFF
--- a/docs/community/software/harborscale.mdx
+++ b/docs/community/software/harborscale.mdx
@@ -7,56 +7,76 @@ sidebar_position: 5
 
 # Harbor Scale Meshtastic Integration
 
-The Meshtastic Integration script allows you to connect your Meshtastic device to Harbor Scale effortlessly. By running this Python code, you can send telemetry data from your Meshtastic device to your Harbor Scale account, where it can be visualized using Grafana.
+The Meshtastic Integration allows you to connect your Meshtastic LoRa mesh to Harbor Scale effortlessly using the **Harbor Lighthouse** agent. Unlike previous methods that required manual Python scripts, the new agent handles device detection, data parsing, and secure transmission automatically.
 
 [To Create an Account - Harbor Scale](https://harborscale.com)
 
 ![Image](https://i.imgur.com/DLqunXx.jpeg)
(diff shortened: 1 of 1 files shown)


rcarteraz:
Sorry, we're not going to recommend people install an executable from an unknown source in our official docs.

PR #2165, opened 2025-12-27
Update LoRa documentation for node ignoring options

RCGV1:
Clarified the note about ignoring nodes in iOS to recommend users to use the newer ignore nodes function.

### DIFF
--- a/docs/configuration/radio/lora.mdx
+++ b/docs/configuration/radio/lora.mdx
@@ -117,7 +117,7 @@ This setting controls the actual hardware frequency at which the radio transmits
 For testing it is useful sometimes to force a node to never listen to particular other nodes (simulating radio out of range). All nodenums listed in the ignore_incoming array will have packets they send dropped on receive (by router.cpp)
 
 :::note
-This option is **deprecated** in the mobile apps; however, the recommended way to ignore nodes is to use the newer option of marking a node as ignored in the apps. This acts the same as the ignore list (blocking packets from the ignore node from being received or relayed), with the added benefit of being able to ignore a lot more nodes.
+This option is **deprecated** in the client apps.  To ignore nodes, long-press on (or click for web client) a node and mark it as ignored. This acts the same as the ignore list (blocking packets from the ignore node from being received or relayed), with the added benefit of being able to ignore more than three nodes.
 :::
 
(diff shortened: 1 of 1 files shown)


docs/configuration/radio/lora.mdx
jschrempp, line 1: Typo:  of beingable to     
Should be:  of being able to

docs/configuration/radio/lora.mdx
```diff
@@ -116,6 +116,10 @@ This setting controls the actual hardware frequency at which the radio transmits
 
 For testing it is useful sometimes to force a node to never listen to particular other nodes (simulating radio out of range). All nodenums listed in the ignore_incoming array will have packets they send dropped on receive (by router.cpp)
```
b8b8, line 117: This has been deprecated from android as well.
The preferred method to block nodes on iOS and Android is to toggle the "ignore" setting on the node list.

pdxlocations:
Should Ignore Incoming Array be removed from the docs entirely?

pdxlocations (changes requested):
Note applies to all clients, not just ios

docs/configuration/radio/lora.mdx
```diff
@@ -117,7 +117,7 @@ This setting controls the actual hardware frequency at which the radio transmits
...
 
 :::note
-This option is not available in iOS; however, the recommended way to ignore nodes is to use the newer option of marking a node as ignored in the apps. This acts the same as the ignore list (blocking packets from the ignore node from being received or relayed), with the added benefit of being able to ignore a lot more nodes.
+This option is **deprecated** in the mobile apps; however, the recommended way to ignore nodes is to use the newer option of marking a node as ignored in the apps. This acts the same as the ignore list (blocking packets from the ignore node from being received or relayed), with the added benefit of being able to ignore a lot more nodes.
```
pdxlocations: How about something like:

This option is **deprecated** in the client apps.  To ignore nodes; long-press on (or click for web client) a node and mark it as ignored. This acts the same as the ignore list (blocking packets from the ignore node from being received or relayed), with the added benefit of being able to ignore more than three nodes.

b8b8: Looks good.

PR #2167, opened 2025-12-29
Significant Updated and Re-write of web-client.mdx

mrdvt92:
Significant Updated and Re-write of web-client.mdx
  - Cleaned up documentation on support between http vs https
  - Updated language to match current application terminology
    - e.g. s/node/device/, s/application/client/, etc.
  - Added Docker example


mrdvt92:
Due to my position, I cannot sign the CLA.  But, my contributions may be considered in the public domain.

PR #2179, opened 2026-01-07
Add Flasher link to navbar and hide the standard external link SVG icon

thebentern:
## What did you change
Added a new top level nav bar link directly to the Web Flasher

## Why did you change it
Streamlined access to the Web Flasher from the Meshtastic home page

## Screenshots

image


rcarteraz:
Only issue I see is mobile. I would recommend we either remove downloads, or not add this (since it’s already in downloads). 
IMG_5390

rcarteraz (approved):

PR #2196, opened 2026-01-16
MUI - Mention Serial Option in BT Programing Mode

### LINKED ISSUE #2065: [Documentation Change Request]: Explain USB OTG support under Meshtastic UI
### Request Details

Currently only Bluetooth mode is explained here when using Meshtastic UI: https://meshtastic.org/docs/software/meshtastic-ui/#bluetooth-programming-mode

However, I heard from somebody that USB OTG also works, if your mobile phone has OTG support of course. Could this be documented?

**Rationale:** In case a user want to program their Meshtastic device (eg. T-Deck) with more advanced settings, currently not available in the UI.

Thanks!

Regards,
Melroy

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

pdxlocations:
This PR adds clarification that serial connections may also be used when in Bluetooth Programming Mode.

Closes #2065


rcarteraz:
This is fine but I also think it would be a good idea to document the unified firmware, which would allow the same capability when in baseui mode.

pdxlocations:
I'm unclear on what you're asking for in this PR.

rcarteraz:
> I'm unclear on what you're asking for in this PR.

The unified firmware, which is all -tft versions, includes both BaseUI and MUI. You can also reboot into BaseUI, and have full API availability similar to BT programming mode. We might want to document that in addition to having the BT programming mode, you can also do these things if you reboot into BaseUI.

pdxlocations:
Aha, yes that would make sense to add here!  I so rarely use my tdeck, I'd forgotten all about this.  Have a look at the changes when you're able.

PR #2201, opened 2026-01-17
Add regulatory document for Belgium in region table

iooner:
Added regulatory document link for Belgium.


pdxlocations:
I'm looking at this page but do not see anything related to LoRa.  Am I missing something?  Thank you.

iooner:
There is nothing specific to LoRa. I have changed the link to the general frequency rules.

pdxlocations:
What value does this link add?  Could there be a more relevant section to link to?

PR #2202, opened 2026-01-18
Update supported devices in getting started guide

Mainstreetjay:
Add RAK3312 Core module to supported devices list.


rcarteraz:
This location isn't meant to be all encompassing. That's what the supported devices section, linked at the top, is for. This is just a brief overview of some supported devices. The 3312 is listed here: https://meshtastic.org/docs/hardware/devices/rak-wireless/wisblock/core-module/?rakcore=RAK3312

Mainstreetjay:
It should be here so people know what they are doing . Totally a lack of knowledge here 


https://meshtastic.org/docs/getting-started/

This is more or less the link given out all over social media, so people can start and one of the things that they can start with is not even listed

PR #2206, opened 2026-01-20
Ru/local user groups and communities

krotesk:
## What did you change

Completely edited the Russian groups. The groups are taken from the Russian telegram group.

## Why did you change it

I decided to update the list of groups.

### Before

Снимок экрана от 2026-01-20 10-30-41

### After

Снимок экрана от 2026-01-20 14-57-32
Снимок экрана от 2026-01-20 14-57-36
Снимок экрана от 2026-01-20 14-57-41
Снимок экрана от 2026-01-20 14-57-45
Снимок экрана от 2026-01-20 14-57-51


rcarteraz:
Updating the display name on this page  does not solve the issue that every one of these groups are in violation of the [trademark guidelines](https://meshtastic.org/docs/legal/licensing-and-trademark/#social-media). Every single one is using Meshtastic in its name in violation of the policy, and every single one is using the logo in violation of the policy. We do not add groups that are not in compliance.

PR #2208, opened 2026-01-21
docs: standardize MCU listing and add device specifications

Luen:
- Make MCU listing consistent for both devices (Nordic nRF52840 with BLE 5.0)
- Add device-specific specifications:
  * Dimensions and weight
  * Battery life (GPS usage)
  * Detailed charging information
  * Waterproof rating details
  * User interface specifications
- Update LoRa transceiver naming to include manufacturer (Semtech)


rcarteraz:
Our docs are not meant to be all encompassing. We give small pertinent info, anything in detail should be obtained from the manufacturer -- which we link.

PR #2212, opened 2026-01-22
Replace biome with oxfmt & oxlint

danditomaso:
## What did you change
I replaced our default formatting/linting tool (Biome) with oxlint and oxfmt. This is an improvement as oxc is compatiable with eslint which is the default linter for JS/TS projects. This will reduce the amount of false positives or failing our pipeline for no reason during our ci/cd 

## Why did you change it
Ben didn't lint it. Also I wanted to unify our linting/formatting set up with the web client


rcarteraz:
@danditomaso The updated homepage has caused merge conflicts with this one. Can you take a look?

danditomaso:
Sure I'll take a look

> @danditomaso The updated homepage has caused merge conflicts with this one. Can you take a look?

PR #2215, opened 2026-01-23
Add Shared Default Key description to Overview > Encryption

eap:
Summary: Document actual value for shared defualt channel key `AQ==`

## What did you change

The introductory section in `about/overview/encryption/index.mdx` includes several inaccuracies that require clarification or expansion.

1) The introduction of the aricle claims "each packet" is encrypted with "AES256-CTR". This is misleading as there's several different paths for encryption.

2) The key `AQ==` is not an actual AES key and instead is an alias for this key in [src/mesh/Channels.h](https://github.com/meshtastic/firmware/blob/develop/src/mesh/Channels.h#L143-L145). 

## Why did you change it

This was a documentation tripwire for me and sent me on an unnecessary quest of researching how to expand a "0x01" key into 256 bits. In fact the key wasn't a 256 bit key nor would expanding it help me since the key is "random" according to th header file. I hope some future tinkerer benefits from the docs being more accurate here.


rcarteraz:
> I hope some future tinkerer benefits from the docs being more accurate here.

Personally, I think all this achieves is to make it even more confusing to people.

eap:
@rcarteraz 
> Personally, I think all this achieves is to make it even more confusing to people.

Can you clarify here? In all documentation there's a balance between breadth and depth, but user docs shouldn't simplify to the point of being misleading. The original statement that the public channel is `encrypted with a simple known key ("AQ==")` is false. `AQ==` is not an AES256 key nor can it be used to derive one.

I could propose eliminating the reference to `AQ==` in this article, although that creates a documentation gap. This _key alias_ is user-facing in the configuration applications; we need to document somewhere that `AQ==` is not a key but is instead an alias to the real key.

If this document on the overview of encryption is the wrong place to document the `AQ==` key alias I'd invite input on where it should be documented.

eap:
I had expected my first change to be the __less__ controversial of the two changes I wanted to make, but as I now anticipate this PR will undergo revisions and deeper discussion. I have added my second change that reworks the introductory paragraph on the basis of _'in for a penny, in for a pound'_.

My overarching goal with this whole thing is to improve the accuracy and correctness of this article while maintaining the "Inverted pyramid" style of information delivery; the opening should **accurately** but superficially cover the topic with increasing depth and detail as one continues reading.

rcarteraz:
> Can you clarify here?

More words does not mean it's easier to understand.

eap:
> More words does not mean it's easier to understand.

I'm trying to be humble and nice here because and I'm earnestly requesting that you to ask questions, make suggestions, or find some way to offer a path forward here.

More word is not necessarily easy to understand, but less words is not necessarily better. Summarizing information at the cost of creating false simplifications is more harmful than having no documentation at all. Calling AQ== an AES256 key is actively hostile to readers.

docs/about/overview/encryption/index.mdx
```diff
@@ -9,16 +9,20 @@ description: "Understand Meshtastic's encryption: optional network-wide AES256 s
...
-
 The device will decrypt the payload before sending it to a client app via BLE, serial, Wi-Fi/Ethernet. For MQTT, you can [specify](/docs/configuration/module/mqtt#encryption-enabled) whether you want to send an encrypted or unencrypted payload.
 
+### Shared Default Key
+
+The default Meshtastic primary channel uses a fixed, well-known AES128 key that is distributed to all Meshtastic devices. This channel exists to allow zero-configuration interoperability between new or unconfigured nodes. This shared key also allows devices to receive and broadcast node info, position, device telemetry, and information from [other modules](/docs/configuration/module/). This public information enriches your client software's understanding of the local mesh and sharing it contributes to the quality of the mesh for other users. If you choose not to use the default channel, your device will participate in the mesh as a relay node, but you will not be able to access public broadcasts or benefit from applications that require intermediate decryption.
```
Jorropo, line 22: > The default Meshtastic primary channel uses a fixed, well-known AES128 key that is distributed to all Meshtastic devices. This channel exists to allow zero-configuration interoperability between new or unconfigured nodes.

We don't need to explain what a public key is more than once, if it isn't clear fix the sentence / paragraph explaining this.

> This public information enriches your client software's understanding of the local mesh and sharing it contributes to the quality of the mesh for other users. If you choose not to use the default channel, your device will participate in the mesh as a relay node, but you will not be able to access public broadcasts or benefit from applications that require intermediate decryption.

We can assume peoples are somewhat familiar with the features.
No need to state the meshtastic features.

docs/about/overview/encryption/index.mdx
```diff
@@ -9,16 +9,20 @@ description: "Understand Meshtastic's encryption: optional network-wide AES256 s
...
+
+The shared key is shown in client applications as `"AQ=="`. This is an alias of the real 16 byte key and is **not** the actual AES key. Using this alias instructs the firmware to use the built-in public channel key. The actual AES128 key used for the public channel is `"1PG7OiApB1nwvP+rz05pAQ=="` (base64 encoded). This key is constant, non-secret, and published in the firmware source and elsewhere. As a result, anyone following local traffic or monitoring via [MQTT](docs/configuration/module/mqtt/) can decrypt messages on this channel.
```
Jorropo, line 24: The first part, (`AQ==` → `1PG7OiApB1nwvP+rz05pAQ==`) is only interesting to peoples writing meshtastic implementations.
Which is not the target of this article. It is expected that you will have to read the code to do so.

---

> This key is constant, non-secret, and published in the firmware source and elsewhere. As a result, anyone following local traffic or monitoring via [MQTT](docs/configuration/module/mqtt/) can decrypt messages on this channel.

This is a round about way to say:
> If you use a public key, everyone can decrypt theses messages.

Which already aims to be covered by:
> By default, you have one primary channel, which is encrypted with a simple known key ("AQ=="), so to use proper encryption, you **must change** this key first, or create a new channel and share it with the ones you want to communicate with. However, without a channel using the default key, communication with devices still using it will not be possible.

No need to repeat it.

docs/about/overview/encryption/index.mdx
```diff
@@ -9,16 +9,20 @@ description: "Understand Meshtastic's encryption: optional network-wide AES256 s
...
 
-By default, you have one primary channel, which is encrypted with a simple known key ("AQ=="), so to use proper encryption, you **must change** this key first, or create a new channel and share it with the ones you want to communicate with. However, without a channel using the default key, communication with devices still using it will not be possible.
+By default, you have one primary channel, which is encrypted with a [known default key](#shared-default-key). The primary channel is used for all periodic broadcasts generated by the device (position, telemetry, traceroutes, etc.). To have private communication, you must create a channel with a private key and share it only with the devices you intend to communicate with.
```
Jorropo, line 14: I think the sentence you are replacing could be improved but your replacement has issues:

> The primary channel is used for all periodic broadcasts generated by the device (position, telemetry, traceroutes, etc.).

*snip* we don't need to state general features of meshtastic here.

---

> By default, you have one primary channel, which is encrypted with a [known default key](#shared-default-key).

We can keep the "lie" about `AQ==`, there are strictly zero security differences between one hardcoded public AES key and an other.

I don't know the history as to why `AQ==` isn't Zexted and used at the key, but this is noise for the point this page conveys.

rcarteraz:
> > More words does not mean it's easier to understand.
> 
> I'm trying to be humble and nice here because and I'm earnestly requesting that you to ask questions, make suggestions, or find some way to offer a path forward here.
> 
> More word is not necessarily easy to understand, but _less words is not necessarily better_. Summarizing information at the cost of creating false simplifications is more harmful than having no documentation at all. Calling `AQ==` an AES256 key is actively hostile to readers.

Based on some of your comments and commit messages, I don't really feel that's the case. Additionally, it’s generally not the best way to introduce yourself as a first-time contributor by making a large change solely because you found something confusing. Engaging in discussion, whether by opening an issue to ask questions or share insights, or by joining the Discord server to do the same, usually works better than submitting PRs based only on your individual experience rather than on what might be best for the project as a whole. Who knows, maybe you’re right and others were confused too, but we can’t know that without having the conversation first.

That said, I’m open to these changes if they pass review by one of the encryption devs who has been asked to take a look.

docs/about/overview/encryption/index.mdx
```diff
@@ -9,16 +9,20 @@ description: "Understand Meshtastic's encryption: optional network-wide AES256 s
 
 ## Explanation
 
-Meshtastic provides AES256-CTR encryption for the payload of each packet when sending via LoRa, with a different key for each [channel](/docs/configuration/radio/channels/). The [packet header](/docs/overview/mesh-algo/#layer-1-unreliable-zero-hop-messaging) is always sent unencrypted, which allows nodes to relay packets they can't decrypt as well. One can disable this by setting a different [rebroadcast mode](/docs/configuration/radio/device#rebroadcast-mode).
+Meshtastic encrypts packet payloads using either a symmetric key or public key depending on the content and destination. [Channel](/docs/configuration/radio/channels/) broadcasts use AES in counter (CTR) mode, with a 128-bit key for the shared default channel and 256-bit keys for private channels. Direct messages use public key cryptography. The [packet header](/docs/overview/mesh-algo/#layer-1-unreliable-zero-hop-messaging) is always sent unencrypted, which allows nodes to relay packets they cannot decrypt. This behavior can be disabled by selecting a different [rebroadcast mode](/docs/configuration/radio/device#rebroadcast-mode).
```
Jorropo, line 12: This whole paragraph isn't great eitherway.

The old version is a bit inaccurate.
Trying to shove PK just makes it worst.

(I think we need two pages.)

---

> One can disable this by setting a different [rebroadcast mode](/docs/configuration/radio/device#rebroadcast-mode).

Needs to be removed, this isn't a page about rebroadcast modes.
The word `allows` already convey the idea that nodes might or might not relay packets.

tekstrand:
FYI @eap since it looks like you're a pretty regular contributor to software in general, you'll find the Meshtastic development community very different than most. e.g. opening a PR first is generally frowned upon here, as @rcarteraz mentioned.

It's weird to get used to.

Jorropo:
Hey, I don't think this page is in a great state for reasons\* imo the most important point is that counter intuitively "AES256 encryption" means nothing as to the actual security implemented, and I regularly came across peoples who were confused and thought meshtastic were a secure messaging platform for your drug empire (/hj).

I've never seen one again seen updating this page, I believe this is thanks to theses two lines:
> ## Is it as secure as Wi-Fi WPA3, HTTPS TLS1.3, or Signal?
> **No**.

If anything, improvements to this page should keep the goals of name dropping AES, CTR & friends and conveying that not all AES256 are equal as quickly as possible.

\*this page was created before 2.5.0 PK DM cryptography was introduced and made sense.
Now after having being updated it contradicts itself in different parts because they talk about groups vs DM encryptions which work completely differently.

I want to try splitting this into two pages, one for DM encryption and one for groups encryption.

Jorropo:
With the goal of strengthening the message I want readers to get after reading this page,
I'll try a PR splitting the two pages and removing things that don't directly relate to the encryption's security model.

eap:
@Jorropo  - Thank you for the review. I'm more than happy to hand this off as I see you want very specific changes.

I would like to respond to this one comment;

> @Jorropo 
> We can keep the "lie" about AQ== there are strictly zero security differences between one hardcoded public AES key and an other.

Readers of this article interested in security implementation will likely observe that AQ== decodes to `0x01` which then invites the interpretation that the key is `0x000....0001` (or the inverse if we assume big endian). Further accentuating the problem is the relative inaccessability of the true value in the codebase; while most cryptographic behaviors are neatly contained in [CryptoEngine](https://github.com/meshtastic/firmware/blob/develop/src/mesh/CryptoEngine.cpp) the default channel key is only found through a reading of `initDefaultChannel` in [Channels.cpp](https://github.com/meshtastic/firmware/blob/develop/src/mesh/Channels.cpp)

The truth is that it is easier to find the true value through Google and various blog posts than in the meshtastic docs or the meshtastic firmware source code.

As these are your docs, I'm not going to argue or dispute your strategic direction on intended audience and it seems I was not the intended audience of this article. __I will request/suggest that the "lie" is not preserved__. It would be preferrable to call that value an alias, or identifier, or to not reference it at all. Any reader interested in crypto-implementation is at least somewhat likely to make the same mistake when the official docs for Meshtastic encryption state that "AQ==" is the key for the public channel.

PR #2223, opened 2026-01-26
Add docs link to links.json

danditomaso:
## What did you change
This fixes the broken link  on the "read docs" button on the homepage. It was removed in a prior commit.

## Why did you change it

Because it was broken.

### DIFF
--- a/src/data/links.json
+++ b/src/data/links.json
@@ -1,6 +1,6 @@
 {
   "getStarted": "/docs/getting-started/",
-  "docs": "/docs/",
+  "docs": "/docs/introduction/",
   "ios": "https://msh.to/ios",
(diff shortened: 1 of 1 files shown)


src/data/links.json
```diff
@@ -1,5 +1,6 @@
 {
   "getStarted": "/docs/getting-started/",
+  "docs": "/docs/",
```
pdxlocations: /docs/ leads to a 404, I think this is what is intended?
```suggestion
  "docs": "/docs/introduction/",
```

danditomaso: Thanks, yeah I originally thought i'd leave it as /docs so we can handle where we wanted the redirect to go. But i'll update it to /docs/introduction.

PR #2226, opened 2026-01-27
Clarifying update to library.mdx

bslade:
## What did you change
I made some small changes clarifying that the example code was using the "meshtastic" Python API package afrom https://python.meshtastic.org/

## Why did you change it

I made the update to make it easier for newbies to understand how to use the example code.


docs/development/python/library.mdx
```diff
@@ -4,7 +4,7 @@ title: Using the Meshtastic Python Library
 sidebar_label: Python Library
 ---
 
-An example using Python 3 code to send a message to the mesh, get and set a radio configuration preference:
+An example using Python 3 code with the ["meshtastic" Python API package](https://python.meshtastic.org) to send a message to the mesh, get and set a radio configuration preference.  The ["meshtastic" Python API package](https://python.meshtastic.org) is typically installed using the command `pip3 install meshtastic`:
```
pdxlocations: Thanks for the addition though I think we can improve this section a bit more.  How about something like:

```suggestion
Below is a Python 3 example using the [Meshtastic Python API](https://python.meshtastic.org/)￼ to send a message over the mesh and to read/update a radio configuration setting. Before running the code, install the library by following the [installation instructions](/docs/software/python/cli/installation/).
```

bslade: Sure, that sounds good to me.

PR #2227, opened 2026-01-27
Add Native Command Line Build Instructions

pdxlocations:
This PR: 
- Adds instructions on installing PIO and building for native without VS Code.
- Removes the reusable block for installing dependencies as these instructions now only appear once.


rcarteraz:
> * Removes the reusable block for installing dependencies as these instructions now only appear once.

Isn't this also referenced on the Linux native devices page for installation?

pdxlocations:
> > * Removes the reusable block for installing dependencies as these instructions now only appear once.
> 
> Isn't this also referenced on the Linux native devices page for installation?

It used to be but was removed in https://github.com/meshtastic/meshtastic/commit/1042f18650b72f41fbd0075ad0b50bc8ea7a17c7#diff-a98d35af51676f7add00c680c0e59617ab493fc7834f402f957af28d3eaabde0L9

rcarteraz:
> > > * Removes the reusable block for installing dependencies as these instructions now only appear once.
> > 
> > 
> > Isn't this also referenced on the Linux native devices page for installation?
> 
> It used to be but was removed in [1042f18#diff-a98d35af51676f7add00c680c0e59617ab493fc7834f402f957af28d3eaabde0L9](https://github.com/meshtastic/meshtastic/commit/1042f18650b72f41fbd0075ad0b50bc8ea7a17c7#diff-a98d35af51676f7add00c680c0e59617ab493fc7834f402f957af28d3eaabde0L9)

Ah. Is there any benefit to removing it? Seems like it would be best to just keep it in case we need to reference it elsewhere in the future versus removing/adding it back.

pdxlocations:
> Is there any benefit to removing it?

Just to avoid unnecessary complexity and so "Edit this page" works to quickly add a dependency.   I don't have strong feelings about it either way.  If you'd like me to leave it as a block I'm totally fine with that too.

rcarteraz:
> > Is there any benefit to removing it?
> 
> Just to avoid unnecessary complexity and so "Edit this page" works to quickly add a dependency. I don't have strong feelings about it either way. If you'd like me to leave it as a block I'm totally fine with that too.

No, you’re good to go with this change just curious. Valid point about the edit this page.

PR #2251, opened 2026-02-07
Update local-groups.mdx, adding Greek group site

giannoug:


pdxlocations (changes requested):
Please choose an original URL and name for your group.  Thank you and happy meshing.

pdxlocations:
Thank you for the updates!

The m-powered logo is free to use and there is no trademark on it that i'm aware of but it's certainly not an original concept.

There were issues when Andy Kirby used mpowered as a domain so I'll let a lead take it from here. @rcarteraz ?

rcarteraz:
> There were issues when Andy Kirby used mpowered as a domain so I'll let a lead take it from here. @rcarteraz ?

To answer the domain concern: not technically, but since it does utilize our logo, we still have a say over how it can be used. If it's being used to simply indicate the community is powered by Meshtastic, that is absolutely fine.

However, I am a little concerned about this specific site's design. Because the "m-powered" logo is the only branding used, and the site has adopted a color scheme and documentation style very similar to our own, it could easily lead to user confusion.

I would request that you add a clear disclaimer stating this is an independent community site, not officially endorsed or sponsored by the Meshtastic project. Please also include a link to the official Meshtastic website somewhere clearly visible, like in your header.

giannoug:
@rcarteraz Thanks for taking the time to review this and explain the concern!

We’ve added a visible header link to the official site as Meshtastic Project, and added a footer note that m-powered.gr is an independent community site and does not belong to the Meshtastic project.

We also changed the site accent color away from the Meshtastic green to make the site feel more distinct. The current logo is only a placeholder too, and we plan to replace it soon.

pdxlocations (approved):

PR #2186, opened 2026-01-10
Add Norway Meshtastic community link

fsaev:

PR #2271, opened 2026-02-21
Add MT Users Norway

Jckf:
Adds links to the Norwegian MT hub site 868.no, and our Facebook group and Discord server.

Supersedes https://github.com/meshtastic/meshtastic/pull/2186


Jckf:
@pdxlocations I can change it to only 1 link, yes. The Facebook group has the MPowered logo as the banner, but Facebook is cropping it. I have asked an admin to upload something else.

PR #2280, opened 2026-02-27
Improve warnings for newcomers transmitting without antenna

AF0BL:
First day exploring meshtastic, noticed this warning on some pages, but not on the first page I landed on

## What did you change
Copied existing caution templates to two "getting started" pages that prompt the user to connect USB.
Added detail to the caution, including mentioning USB. 

## Why did you change it
Would hate for newcomers to make an easy mistake


pdxlocations:
I'm not a semtec engineer but while good practice, I've been told that the sx1262 has protection built in to prevent damage from high SWR.  I'm not sure the warning needs to be prominent on multiple pages.  Just my uneducated opinion though.

AF0BL:
@rcarteraz Hi, why did you close this?

PR #2285, opened 2026-03-02
Add LONG_TURBO case to modem preset switch

ae5au:
LONG_TURBO was half-way added in a previous commit

## What did you change
Completed adding LONG_TURBO to the Frequency Calculator

## Why did you change it
It was missing/incomplete


ae5au:
#2195 added the other half but UI is broken 
image

PR #2287, opened 2026-03-03
Add information about the Meshtastic Chirp and Packet

pdxlocations:
This PR creates new sections on the Overview Page to further clarify how Meshtastic works
- The LoRa Chirp - Describes how information is transmitted in a LoRa Chirp for Meshtastic
- Meshtastic Packet Over the Air - Breakdown of packet components
- The MeshPacket - Describes how MeshPacket is used within Firmware

[preview](https://meshtastic-git-fork-pdxlocations-lora-chirp-meshtastic.vercel.app/docs/overview/#the-lora-chirp)


docs/about/overview/index.mdx
jschrempp, line 1: This is great!
You say, "Only nodes with the same Channel name and encryption key can decrypt and interpret the contents."
My understanding is that nodes which have any channel configured with the correct encryption key can decrypt the packet data no matter what Channel name is used. 
I'm no expert in this, but that's what I was told.

pdxlocations: Thanks!  I've been confused on this point as well.  I think I tested this a few years ago and found that the channel name mattered, I'm not sure if that was client-side or in the firmware.

jschrempp: The way it was explained to me is that a receiving device could try all the encryption keys it has against each packet, but that would take time. Instead, the packets have a hash of the channel name and encryption key and a bit of that is in the unencrypted header as a "hint". A receiving node uses this hint to decide which encryption key to try. This saves time. However, a receiving node could ignore the hint and try each key.

ianmcorvidae: The channel name doesn't have to match, but the channel _hash_ needs to in order for your node to attempt decryption, and the channel name goes into that hash. So while it's possible to have collisions and while it might be possible to make custom firmware that tries every key on every packet, this isn't how it works in general.

docs/about/overview/index.mdx
```diff
@@ -26,3 +26,136 @@ Sitting on top of this radio mesh are Channels. A logical mesh is formed by a Ch
...
+   A Cyclic Redundancy Check to detect transmission errors.
+
+Each of these components is transmitted as a sequence of chirps. The spreading factor determines how long each chirp lasts, and the bandwidth determines how wide the frequency sweep is. Higher spreading factors increase range and robustness but reduce data rate.
+
+In Meshtastic, the LoRa chirp carries an encrypted protobuf packet inside the payload section, which includes routing metadata such as hop limit, packet ID, and channel information.
+
+
+## The MeshPacket
+
+The MeshPacket is the primary application-layer structure used by Meshtastic to transport data across the radio mesh. It is serialized using Protocol Buffers (protobuf). Only the application payload is encrypted before transmission. The routing fields—such as from, to, packet ID, hop limit, and channel index remain unencrypted so nodes can correctly forward the packet across the mesh.
```
ianmcorvidae: I don't know how important this is to note, but the header part of the MeshPacket isn't actually a protobuf over the air, but instead is a fixed-width static data structure; only the Data part is actually sent as an encrypted protobuf. Basically, the MeshPacket is only a concept used within firmware and over alternate transports like UDP and MQTT.

docs/about/overview/index.mdx
```diff
@@ -26,3 +26,136 @@ Sitting on top of this radio mesh are Channels. A logical mesh is formed by a Ch
...
+
+6. **Payload (Encrypted)**  
+   The actual application data. This may be:
+   - A text message
+   - Position information (GPS)
+   - Telemetry data (battery, voltage, etc.)
+   - Node info updates
+   - Administrative messages
+
+Before transmission, the MeshPacket payload is encrypted using the Channel’s shared key. Only nodes with the same Channel name and encryption key can decrypt and interpret the contents.
```
ianmcorvidae: As mentioned in the other comment thread, this is largely correct -- nodes don't try all keys on all messages, there's just a chance of hash collisions that might make differing channel names happen to work.

If you wanted to be really precise, you could say something like: Nodes use a value derived from the channel's name and encryption key (a hash) to identify channels over the air, so nodes will only decrypt packets where one of their configured channels matches the hash sent over the air and the matching channel's encryption key matches the one used by the sender. Therefore, with some exceptions that can usually be ignored, nodes will only decrypt packets sent by nodes using an exactly matching channel (both name and encryption key).

I'm guessing somebody could write that a little less wonky than I did though. You could probably also keep it as it is and just add a footnote that says something like "technically there are a handful of cases where this isn't true", since this isn't really aimed at expert users.

docs/about/overview/index.mdx
```diff
@@ -26,3 +26,136 @@ Sitting on top of this radio mesh are Channels. A logical mesh is formed by a Ch
...
+
+On reception, a node:
+- Verifies it has not seen the Packet ID before
+- Decrypts the payload (if it has the correct Channel key)
+- Delivers the decoded message to the connected client app
+- Rebroadcasts the packet if the hop limit allows
```
ianmcorvidae: Another possibly overly-pedantic suggestion:

```suggestion
- Rebroadcasts the packet if the node's configuration and the packet's hop limit allows
```

docs/about/overview/index.mdx
```diff
@@ -26,3 +26,136 @@ Sitting on top of this radio mesh are Channels. A logical mesh is formed by a Ch
...
+3. **To (Destination Node ID)**  
+   The destination node ID. This may be a specific node, or a broadcast address (4294967295).
+
+4. **Hop Limit**  
+   The maximum number of times the packet may be rebroadcast. Each node that retransmits the packet decrements this value. When it reaches zero, the packet is no longer forwarded.
+
+5. **Channel**  
+   Indicates which configured Channel the packet belongs to. This determines which encryption key is used.  This channel number is created by using a hash of the channel name and it's PSK and does not correlate with the client channel idexes.
+
+6. **Payload (Encrypted)**
```
ianmcorvidae: Technically, as you note below, the encrypted part is really a wrapper around the payload plus some more metadata, specifically the portnum, bitfield, and some additional things like response and reply IDs (yes, those are two different things), the emoji flag, etc. I'm not sure if this is important to distinguish here, though.

ianmcorvidae (commented):
I don't know if any of this is helpful but I got nerd-sniped with the pedantry :joy:

Take or leave anything here as you deem best :)

docs/about/overview/index.mdx
pdxlocations: I'd like to keep this "simplified" and there are also situations where having a matching channel and key _won't_ be decoded, such as using a different frequency slot.  I mention the hash lower down though mostly as a way to differentiate the ota channel from the client channel.  I'm open to improvements here of course.

docs/about/overview/index.mdx
pdxlocations: I'm a bit hazy on the term header to be honest.  LoRa chirps have a header (cr/etc), then Meshtastic has an 8 bit packet header [described here](https://meshtastic.org/docs/overview/mesh-algo/#packet-header-flags), then the packet has an unencrypted header (from/to/chan/id/etc).

I definitely don't want to use the term meshpacket incorrectly (though it does sound nice). Perhaps I should change references in this document to "data"? But there's another ambiguous term if there ever was one.

docs/about/overview/index.mdx
pdxlocations: I was about to commit to this, but then had the thought that we're speaking in general terms here, defaults if you will.

docs/about/overview/index.mdx
ianmcorvidae: Yeah, for sure. So to clarify somewhat on the headers:

LoRa itself has a header, as you note. That has nothing meshtastic-specific and is stuff like CR, again as you noted.

After that comes the payload from the LoRa perspective. This includes all the data that's present in a MeshPacket protobuf, but it's not encoded strictly as a MeshPacket protobuf. Instead, it's split into two parts. We'll ignore the second part for now, because that's the encrypted part of the packet. The first part is the 16-byte meshtastic header. That's this part of the table above where you linked on the documentation page:

image

In other words, that header contains the packet destination (to), sender (from), packet ID, flags, channel hash, next hop byte, and relay node byte. The flags byte of this header contains what you linked -- hop_limit, want_ack, via_mqtt, and hop_start. So in total, the 16-byte meshtastic header is, in order and not encoded as protobuf (that is, these are just directly in this order with nothing between them or denoting the structure explicitly):

* 4 little-endian bytes representing the destination
* 4 little-endian bytes representing the sender
* 4 little-endian bytes representing the packet ID
* 3 bits representing the hop limit
* 1 bit for want_ack
* 1 bit for via_mqtt
* 3 bits for the hop_start value
* 1 byte for next hop
* 1 byte for relay node

or, tl;dr, there's two headers: The LoRa one that denotes LoRa stuff, then the Meshtastic one that denotes Meshtastic stuff. After that is the encrypted part of the packet.

Of these three parts, only the encrypted part of the packet is actually encoded as protobuf when transmitted over LoRa. However, when sent over the client API, MQTT, or UDP, the Meshtastic header is encoded as protobuf for ease of use.

I think if you wanted to be precise about what a MeshPacket is, you might change what you're saying here to something like:

> The MeshPacket is the primary application-layer structure used by Meshtastic to transport data across the radio mesh. For direct user uses, it is serialized using Protocol Buffers (protobuf). Over the air, part of the data is split out into a more space-efficient but less flexible header. Only the remainder, the application payload, is encrypted before transmission. The routing fields such as from, to, packet ID, hop limit, and channel information are sent in the header and remain unencrypted so nodes can correctly forward the packet across the mesh.

As usual I have no idea if this makes any sense how I'm putting it :sweat_smile:. But hopefully this gives you enough information to simplify it into something useful for your target audience here.

pdxlocations:
@ianmcorvidae Thanks again for your wisdom.  I think I've incorporated your notes into this PR.  I may have missed or misinterpreted something, but I'm hoping this is steps in the right direction.

rcarteraz (approved):

PR #2334, opened 2026-04-19
Update NPM modules (Docusaurus 3.10.0)

vidplace7:
Update NPM modules.

Most notably [Docusaurus 3.10.0](https://github.com/facebook/docusaurus/releases/tag/v3.10.0)

Also add dependabot definition for NPM modules so this doesn't happen again.

Should fix `quality` checks (complaining about oxlint)


danditomaso (approved):
Looks good. Thanks for getting this updated!

Issue #2321, opened 2026-04-04
[Documentation Change Request]: social media links on front page don't wrap on small screens


NomDeTom:
### Request Details

As title 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

rcarteraz:
Fixed by #2335

PR #2335, opened 2026-04-20
Fix issue #2321 mobile social links - also fix copyright footer visibility 

basilleaf:
## What did you change

This changes mobile social links so they wrap to 2 lines on mobile viewport size, and also makes them not sticky in mobile view. It also fixes the visibility of the Copyright footer. 

## Why did you change it

Wrapping the social links to 2 lines fixes issue #2321 but it also makes this section taller, which on a mobile device since that section is sticky as you scroll, being taller is not ideal, it can overlap with the `Get Started` button etc. So this change also makes it not sticky and just have it in the document at the bottom of page. 

Also noticed the Copyright footer was not displayed at all, so the 2nd commit fixes that to display the copyright footer that was already there - looked to be a z-index issue, being overlapped by the site background

### Before

https://github.com/user-attachments/assets/e35bfe2d-8bf6-4164-95e6-24068353e362

### After

https://github.com/user-attachments/assets/fdad40e4-7cbd-4e52-a19d-6456836c1994

Desktop is unchanged: 

image


rcarteraz:
@basilleaf This looks good! Going to have our web guy take a look.

danditomaso:
@basilleaf Thank you for this PR. I am the orginal author of the Meshtastic.org site and I'm sure I was rushing while completing the narrow mobile view, so thank you for submitting these fixes.

danditomaso (approved):
LGTM

basilleaf:
Thanks for merging! 🎉

PR #2339, opened 2026-04-24
Translate introduction.mdx to German

nachbarschaft-weidenau-de:
## What did you change

We translated the page to German. 

Our understanding is, that https://meshtastic.org/de/docs/introduction/ will display the currently text in english in german. 

## Why did you change it

Contributing to the project. German texts might be easier to understand for Germans in our German neighbourhood.

### DIFF
--- a/docs/about/introduction.mdx
+++ b/docs/about/introduction.mdx
@@ -1,74 +1,74 @@
 ---
-title: Introduction
-sidebar_label: Introduction
-slug: /introduction
+title: Einführung
+sidebar_label: Einführung
+slug: /einfuehrung
 sidebar_position: 1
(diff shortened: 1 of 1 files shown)


rcarteraz:
Translations are handled via crowdin, and we actually have 100% German translations. What you're attempting to do completely overwrites the English docs. We do, however, have an issue with the crowding plugin not properly integrating the translations which is causing problems. We're trying to get that sorted. Thank you for your contribution, but I will be closing this at this time.

nachbarschaft-weidenau-de:
Oh, that  s cool. all good.

https://meshtastic.org/**de**/docs/introduction/ seemed to not be the English docs, but then again, you know best.

Can I have a look at the crowdin translations or an estimation on the plugin-fix, please?

PR #2379, opened 2026-05-05
Update description of User Names

richb-hanover:
## What did you change
Updating the description of User Names

## Why did you change it
As a newcomer to Meshtastic, I wasn't totally clear on the meanings of the documentation. Please verify that I've got it right (or I'll be happy to fix it.)

Thanks!


docs/configuration/radio/user.mdx
```diff
@@ -9,23 +9,29 @@ import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 import { Icon } from "@iconify/react";
 
-The user config options are: Long Name, Short Name, Is Licensed, and Is Unmessageable. User config uses an admin message sending a `User` protobuf.
+The User Config options are: Long Name, Short Name, Is Licensed, and Is Unmessageable. User config uses an admin message sending a `User` protobuf.
+
+In all the names below, #### represents the last four digits of the MAC adress, and each emoji consumes four bytes of the string.
```
rcarteraz:
```suggestion
```

docs/configuration/radio/user.mdx
```diff
@@ -9,23 +9,29 @@ import Tabs from "@theme/Tabs";
...
 
-### Short Name
+A personalized identifier for your device.
+
+Defaults to the `ShortName`. This can be changed to any string, including emoji, up to 36 bytes.
+
+If you are a licensed ham operator and have enabled `IsLicensed`, you should set the Long Name to your ham operator call sign.
 
-A personalized short identifier for your device.
+### BLE Connection Name
```
rcarteraz:
```suggestion
```

rcarteraz: This is not a config and will only cause confusion.

docs/configuration/radio/user.mdx
```diff
@@ -9,23 +9,29 @@ import Tabs from "@theme/Tabs";
...
 
-Auto-generated by default.
+The name assigned to the Bluetooth connection. Always set to `ShortName_####`
```
rcarteraz:
```suggestion
```

rcarteraz (changes requested):
Why are we swapping long name and short name? That is an unnecessary change. I've also made suggestions for other unnecessary changes. Please update these and then I'll review the verbiage you've supplied for long and short name.

richb-hanover:
Hi! Thanks for your comments. I made this PR so the docs can supply information that a newcomer (like me) can use right away, with out the need to make several experiments to find the limits of the system. 

I would encourage you to check what I've written for correctness - as I said, I'm a newcomer and have only discovered this by actually experimenting with the radio. Alternatively, point me to places in the existing documentation where this information is written. Now to your specific comments:

- I put ShortName first because then the LongName can say, "Defaults to ShortName" which had already been defined. 

- I added the sentence about `####`, emojis, and the length so that they are documented here. This helps the reader know the constraints/limits on the values described below. 

- I added the BLE Connection Name because it appears in the GUI (perhaps on the phone app). If it's documented elsewhere, please give me a link. I agree that I could change the sentence to indicate that this is set automatically by the radio, and cannot be changed.

- Finally, please check that what I've written is correct. 

Thanks again

rcarteraz:
> * I added the sentence about `####`, emojis, and the length so that they are documented here. This helps the reader know the constraints/limits on the values described below.



> Hi! Thanks for your comments. I made this PR so the docs can supply information that a newcomer (like me) can use right away, with out the need to make several experiments to find the limits of the system.
> 
> I would encourage you to check what I've written for correctness - as I said, I'm a newcomer and have only discovered this by actually experimenting with the radio. Alternatively, point me to places in the existing documentation where this information is written. Now to your specific comments:
> 
> * I put ShortName first because then the LongName can say, "Defaults to ShortName" which had already been defined.
> * I added the sentence about `####`, emojis, and the length so that they are documented here. This helps the reader know the constraints/limits on the values described below.
> * I added the BLE Connection Name because it appears in the GUI (perhaps on the phone app). If it's documented elsewhere, please give me a link. I agree that I could change the sentence to indicate that this is set automatically by the radio, and cannot be changed.
> * Finally, please check that what I've written is correct.
> 
> Thanks again

Hi! Thank you for the PR and for taking the initiative to help improve the onboarding experience for newcomers. I really appreciate your perspective.

To address your point about checking for correctness: while ensuring technical accuracy is absolutely vital, correctness alone isn't the only goal for our documentation. For documentation to be effective, information must also be placed contextually where users expect it, formatted for general comprehension, and broadly applicable. Something I always advocate for, is we have to be careful not to restructure pages based on a single user's initial experience; instead, we look for recurring feedback trends to inform how we evolve the docs.

As I mentioned previously, I left some suggestions regarding the formatting first. In my opinion, the proposed structural changes don't currently make the page easier for new people to comprehend. Once we get the formatting and placement aligned, I will gladly review the specific verbiage for the Short and Long names.

Regarding your specific points:

1. The Long Name does not default to the Short Name. It defaults to "Meshtastic_####" (where the last 4 digits are the device's MAC address). The default Short Name is simply the last 4 digits of the MAC address. Saying the Long Name defaults to the Short Name is technically inaccurate and could confuse a new user. If we want to clarify this, we should spell out exactly how the MAC address dictates the default for both, which I have added suggestions for.
2. The sentence you added regarding ####, emojis, and character length is not entirely correct in a broad sense, nor does it apply to the entire config page. To be helpful, these constraints need to be moved and attached directly to the specific Long Name and Short Name sections rather than sitting as a general rule.
3. This is client-specific (iOS/Android/Web) and belongs in those respective documentation sections, not the general user config page. For example, it could be added to the documentation here: https://meshtastic.org/docs/software/android/usage/#connecting-to-your-radio. Adding it to this specific config page, even with a disclaimer, will create confusion in terms of it being a setting that can be changed (which it is not). 

I'm happy to review the PR further once we've addressed the structural and placement changes mentioned above. From there, I look forward to collaborating with you on fine-tuning the exact verbiage for the long and short names to ensure they are as clear and helpful to readers as possible.

roberthadow:
Hello @rcarteraz --

> To address your point about checking for correctness: while ensuring technical accuracy is absolutely vital, correctness alone isn't the only goal for our documentation. For documentation to be effective, information must also be placed **_contextually where users expect it, formatted for general comprehension, and broadly applicable_**. Something I always advocate for, is we have to be careful not to restructure pages based on a single user's initial experience; instead, we look for recurring feedback trends to inform how we evolve the docs.

As to restructuring pages, I suggest we take a winner, make it a little bit better, then use the same topics, form and feel for other less developed bits of documentation.  

1. Android Application Usage -- best of breed. Replicate it for other interfaces
2. Apple Application Usage
3. Web Application Usage
4. -- and to extent it makes sense, Pythod CLI application usage

richb-hanover:
Here's an update to the Short & Long Names. You are correct about the BLE Connection Name: they are different between iOS and Android. (I'll ask why in a separate message.) 

I think this addresses the definitions of Short Name and Long Name, unless I got something wrong. Thanks.

garthvh:
> Here's an update to the Short & Long Names. You are correct about the BLE Connection Name: they are different between iOS and Android. (I'll ask why in a separate message.)
> 
> I think this addresses the definitions of Short Name and Long Name, unless I got something wrong. Thanks.

This value comes from the firmware, it is the same on iOS and Android

docs/configuration/radio/user.mdx
```diff
@@ -11,28 +11,22 @@ import { Icon } from "@iconify/react";
...
-In all the names below, #### represents the last four digits of the MAC adress, and each emoji consumes four bytes of the string.
-
 ## User Config Values
 
 ### Short Name
 
 The short name for your device.
 
-Defaults to `Meshtastic ####`. Can be changed to any four-byte value or a single emoji (since emojis consume four characters).
+Defaults to `####`, where `####` is the last four characters of the device's MAC address. This can be changed to any four-byte value or a single emoji (since emojis consume four characters).
```
rcarteraz, line 20: ```a single emoji (since emojis consume four characters).```

This isn't entirely correct—not all emojis are four bytes. Many common emojis (like ⚠️) use 3 bytes, while complex ones (like 👨‍👩‍👧‍👦) can use over 20 bytes.

It would be more accurate to say we accept any emoji up to 4 bytes in size.

richb-hanover:
@rcarteraz Thanks for that correction. (I knew emoji were tricky with multiple numbers of bytes...)

I have been impressed by this updated documentation page, so I am going to close this PR. https://meshtastic.github.io/Meshtastic-Apple/user/nodes.html

PR #2425, opened 2026-05-26
Update local-groups.mdx

NebraskaMesh:
Added NebraskaMesh Local group webpage

## What did you change
Added NebraskaMesh as a local group.

## Why did you change it
NebraskaMesh has updated their webpage.

https://www.nebraskamesh.net/


pdxlocations (changes requested):
The linked site is not appear to be relevant to Meshtastic.

PR #2581, opened 2026-08-03
Add Heltec LoRa32 V4-R8 documentation

### LINKED ISSUE #2548: heltec v4
New heltec v4 now using esp32 s3 r8 only version available from meshtastic is esp32 s3 r3 both of my new heltec chips have the new r8 version has my ram when I try to flash meshtastic it fails heltec web site just updated there info showing the changes

dzienisz:
Adds documentation for the Heltec LoRa32 V4-R8 variant, which uses a different ESP32-S3R8 chip and pinout than the original V4.

Changes:
- New V4-R8 tab on the HELTEC LoRa 32 page with specs, pin map, and firmware filenames.
- Warning that V4-R8 requires its own firmware and is not pin-compatible with the original V4.
- Updated hardware index tables to include the V4-R8 row.
- Added the official Heltec V4-R8 pinmap image.

Fixes meshtastic/meshtastic#2548

## Summary by CodeRabbit

* **Documentation**
  * Added support documentation for the HELTEC LoRa32 V4-R8.
  * Documented its hardware specifications, connectivity, display variants, pin mapping, resources, and firmware requirements.
  * Added the device to supported hardware listings and clarified that it requires firmware separate from the original LoRa32 V4.


rcarteraz:
We're still waiting for this hardware so I'm drafting this until that is received.

PR #2599, opened 2026-08-13
Firmware docs migrate

NomDeTom:
Per @caveman99 , moving the code documentation out of the firmware and into docs,

## Summary by CodeRabbit

* **Documentation**
  * Added user guides for the Mesh Beacon and Traffic Management modules, including configuration, behavior, compatibility, and client support.
  * Added integration guidance for Mesh Beacon clients and region-specific modem preset compatibility.
  * Added reference documentation covering Mesh Beacon, traffic management, node information storage, hop scaling, and related firmware behavior.
  * Updated the module documentation index with links to the new guides.

PR #2600, opened 2026-08-13
docs: move the firmware design docs to the documentation site

NomDeTom:
Per @caveman99, this moves the code documentation out of the firmware repo and into the documentation site.

Supersedes #2599, which was opened from a fork. Same commits, rebuilt on a branch in this repo so the Vercel preview can build without a team authorization step. All review feedback from that PR is already applied here.

The three Mesh Beacon pages have been **split out into their own branch and pull request**. They needed a substantial rewrite for the frequency-slot and by-value channel work in meshtastic/protobufs#1049, and holding this migration behind a firmware change that has not landed helps nobody.

## What this adds

Six pages, all new — no existing page is modified except one link line in the module index.

**Module usage** (`docs/configuration/module/`)

- `traffic-management.mdx` — position deduplication, rate limiting, unknown-packet filtering and NodeInfo direct responses, including the 2.7.x settings surface for devices still on it.

**Client integration** (`docs/development/device/`)

- `region-preset-compatibility.mdx` — `FromRadio.region_presets`, its semantics, and the rules a decoder has to follow.

**Firmware internals** (`docs/development/reference/`)

- `traffic-management-internals.mdx` — the NodeInfo direct-serve path, the throttle bounding it, and the modular tick clocks behind its caches.
- `node-info-stores.mdx` — the hot store, warm tier and two TMM caches: what each holds, capacities per memory class, and which lookup consults which.
- `hop-scaling-internals.mdx` — how the sampled 512-byte histogram estimates mesh size and picks a hop limit for routine broadcasts.
## Why

These documents lived in the firmware tree, where they were invisible to anyone not reading source, and drifted from the code because nothing prompted an update when behaviour changed. Moving them here puts them next to the settings they describe and in front of the people configuring devices and writing client apps.

Rather than transfer them verbatim, each page was checked against current `develop` and corrected where it had drifted. The hop scaling page in particular: its source PR description had the wrong log tag, referenced config flags that no longer exist, described politeness as a multiplier when it is now an interpolation between a floor and a ceiling, and omitted both the flash persistence and the RF-only sampling gate. Its example log lines were also arithmetically impossible — every scaled figure must be a multiple of the filtering denominator, and most weren't.

Every page carries the firmware version it describes, so a reader on 2.7.x can tell what applies.

## Screenshots

Not applicable in the usual sense — these are new pages rather than visual changes to existing ones, so there is no before state. The scenario charts on the hop scaling page ship as paired light/dark SVGs through `ThemedImage`, generated from the same hop-walk the firmware runs so they can't disagree with the algorithm; each palette was checked for contrast against its own surface rather than flipped automatically.

## Verification

- `pnpm lint:mdx` passes.
- Dev server compiles clean, all six routes return 200.
- Chart geometry was verified by decoding bar widths back through the scale and comparing against the source data.
- Firmware claims were checked against `develop` at the source, not from the original documents.

The full production build (`onBrokenLinks: "throw"`) has not been run locally — it exceeds the memory available on my machine — so CI is the first real check of cross-page links.

## Summary by CodeRabbit

- **Documentation**
  - Added documentation for the Traffic Management module, including configuration, filtering, rate limits, role-based behavior, CLI commands, and firmware compatibility.
  - Added developer guidance for region preset compatibility in firmware 2.8.
  - Added reference material covering hop scaling, node and traffic-state stores, and Traffic Management internals.
  - Documented packet handling, caching, persistence, trust rules, throttling, and related implementation behavior.


rcarteraz:
Ongoing work with traffic management and hop scaling, converting back to draft for now.

PR #2619, opened 2026-08-26
docs: separate device display units from app locale units

NomDeTom:
Display config's Units setting only drives the radio's own screen, but it's the first thing people find when the app shows the wrong units.

- Display config: note that the apps follow the phone locale instead.
- Android Units & Locale: new section on the device-screen settings, plus the regions Android treats as imperial (English UK) and phones with no Regional preferences menu.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that preferred display units affect only the device screen.
  * Explained that apps use the client’s locale, such as phone, desktop, or browser settings, and are not affected by this preference.

### DIFF
--- a/docs/configuration/radio/display.mdx
+++ b/docs/configuration/radio/display.mdx
@@ -45,6 +45,8 @@ Acceptable values:
 
 Switch between `METRIC` (default) and `IMPERIAL` units
 
+This setting applies to the screen on the device only. The apps take their units from the phone locale — see [Android](/docs/software/android/user/units-and-locale) and [Apple](/docs/software/apple/user/units-and-locale) — and are unaffected by it.
+
 ### Flip Screen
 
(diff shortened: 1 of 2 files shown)


rcarteraz:
@NomDeTom The android docs are auto-generated and synced from the Android repo. You'll want to work with @jamesarich to update that.

NomDeTom:
@rcarteraz Makes sense, thanks. Reverted the direct edit to `docs/software/android/user/units-and-locale.md` here since it gets clobbered by the next sync-android-docs run anyway — kept only the `display.mdx` note, which isn't a synced file.

The actual content now lives in meshtastic/Meshtastic-Android#6906 against the source (`docs/en/user/units-and-locale.md`), cc @jamesarich. That PR also folds in Copilot's two review comments from here (the "Display Fahrenheit" setting name, and widening "phone locale" to cover Desktop/web too).

rcarteraz (approved):

PR #2627, opened 2026-08-30
feat(FrequencyCalculator): sync region and preset data for 2.8

vidplace7:
## What did you change

Brought `src/components/tools/FrequencyCalculator.tsx` back in sync with the region table and slot math in [firmware `develop`](https://github.com/meshtastic/firmware/blob/develop/src/mesh/RadioInterface.cpp). Scoped strictly to the calculator component — no region tables or other docs were touched.

**New regions (9):** `EU_866`, `EU_N_868`, and the ITU amateur bands `ITU1_2M`, `ITU2_2M`, `ITU3_2M`, `ITU2_125CM`, `ITU1_70CM`, `ITU2_70CM`, `ITU3_70CM`.

**New modem presets (7):** `MEDIUM_TURBO`, `LITE_FAST`, `LITE_SLOW`, `NARROW_FAST`, `NARROW_SLOW`, `TINY_FAST`, `TINY_SLOW`.

**Corrected slot math** — ported from `RadioInterface::applyModemConfig`. The old version ignored padding and used `floor`:

```
slotWidth = spacing + padding * 2 + bw / 1000
numSlots  = round((freqEnd - freqStart + spacing) / slotWidth)
freq      = freqStart + bw / 2000 + padding + slot * slotWidth
```

**Region profiles** replace the flat per-region `spacing` field, carrying the preset list, spacing, padding and `licensedOnly` (`PROFILE_STD`, `EU868`, `UNDEF`, `LITE`, `NARROW`, `HAM_20KHZ`, `HAM_100KHZ`). The preset dropdown now only offers presets the selected region actually supports, and ham regions show a license note.

**Wide-LoRa bandwidths** — `LORA_24` now uses the 2.4 GHz variants (LongFast 812.5 kHz, not 250), changing its slot count from 102 to 103.

**Override slots** — regions can pin a default slot (`EU_N_868` → 1, `ITU2_70CM` → 137, …) instead of hashing the channel name.

**EU trio behaviour** — `EU_868`/`EU_866`/`EU_N_868` are offered the superset preset list the firmware advertises to clients; picking a sibling's preset switches the region to match, with an inline note explaining the switch. Changing the region instead clamps the preset to that region's default, so all three remain reachable.

Two changes beyond a literal port, both inside the component:

- Dropped the `Types.ChannelNumber` import — it is the 0–7 message-channel enum, the wrong type for a slot index that now reaches 300. This removes the component's only `@meshtastic/core` dependency.
- Replaced the two `useEffect` + `setState` passes with values derived during render. The old code set the slot twice on every change, and oxlint flags the pattern (`react(set-state-in-effect)`).

## Why did you change it

The calculator produced wrong slot counts and frequencies for any region that uses channel spacing or padding, and had no entry at all for the regions and presets added to the firmware since 2.7. Anyone using it to pick a slot on the new EU or amateur bands would have got numbers the firmware disagrees with.

## Verification

Every value the firmware documents in its region table comments now matches:

| Region | Preset | Slots | Default slot | Frequency |
| --- | --- | --- | --- | --- |
| US | LongFast | 104 | 20 | 906.875 |
| EU_868 | LongFast | 1 | 1 | 869.525 |
| EU_866 | LiteFast | 4 | 2 | 866.3 |
| EU_N_868 | NarrowSlow | 3 | 1 | 869.442 |
| ITU1_2M | TinyFast | 100 | 26 | 144.510 |
| ITU2_2M | TinyFast | 200 | 51 | 145.010 |
| ITU3_2M | TinyFast | 200 | 33 | 144.650 |
| ITU2_125CM | NarrowSlow | 50 | 37 | 223.650 |
| ITU1_70CM | NarrowSlow | 100 | 37 | 433.650 |
| ITU2_70CM | NarrowSlow | 300 | 137 | 433.650 |
| ITU3_70CM | NarrowSlow | 200 | 37 | 433.650 |

`EU_866` yields the four channel centres named in the firmware comment (865.7 / 866.3 / 866.9 / 867.5) and `EU_N_868` the three at 869.442 / 869.525 / 869.608.

`oxlint` and `oxfmt` are clean, and the component server-renders without error.

## Note for reviewers

The region tables in `docs/about/overview/radio-settings.mdx` still describe the pre-`develop` band plan and do not mention the new EU or amateur regions, so they are out of step with the calculator until updated separately. Left alone here deliberately to keep this PR scoped to the component.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added expanded frequency presets, including Lite, Narrow, Tiny, and ham-radio options.
  - Added support for wider LoRa bandwidth configurations.
  - Added automatic switching between compatible EU frequency bands.
  - Added licensed-band warnings and clearer swap notifications.

- **Improvements**
  - Region changes now select a valid preset automatically.
  - Frequency slots and channel values update reliably with the selected region and preset.
  - Only compatible presets and selectable regions are shown.
  - Region selection now appears before preset selection.


rcarteraz:
Drafting since part of 2.8

rcarteraz:
This seems to change from pulling from the protobuf to hardcoding the presets. Is there a reason for that?

vidplace7:
> This seems to change from pulling from the protobuf to hardcoding the presets. Is there a reason for that?

The presets (and regions) are simply enums in the protobufs, we are continuing to reference them from `@meshtastic/protobufs` (npm package). The actual logic for channel spacing, regional frequency overrides, auto-switching for EU regions, etc all lives in firmware (not protobufs) and had to be re-implemented here.
You'll find the previous version of the calculator also hardcodes quite a bit (everything that is firmware-only).

rcarteraz:
> You'll find the previous version of the calculator also hardcodes quite a bit (everything that is firmware-only).

I'm aware, I was the one that actually had the idea for this and worked with Sacha to get it created. 

As for my question, it was actually a misunderstanding on my part for not reading the change entirely. You just created an alias is all.

rcarteraz (approved):

NomDeTom (approved):

PR #2643, opened 2026-09-03
docs: add the Mesh Beacon module documentation

NomDeTom:
Split out of #2600, which carries the rest of the firmware documentation migration. These three pages needed a substantial rewrite rather than a transfer, and holding the other six behind that helps nobody.

**Draft until the schema lands.** These pages describe `MeshBeaconConfig` and `MeshBeacon` as of meshtastic/protobufs#1049, with the firmware side in meshtastic/firmware#11662. Neither is merged yet, so parts of this document behaviour that does not exist on `master` today.

## What this adds

Three pages, all new; no existing page is modified except one link line in the module index.

- `docs/configuration/module/mesh-beacon.mdx` — the settings, in a user's terms: what a beacon advertises, where it is transmitted, the flags bitfield, legacy split, and what happens when you receive one.
- `docs/development/device/mesh-beacon-client-interface.mdx` — the client-side specification: the wire message, capability detection, config read/write with the full sanitisation table, receiving beacons, and the offer trust model.
- `docs/development/reference/mesh-beacon-internals.mdx` — how the firmware actually behaves: the two module roles, zero-hop stamping, per-packet radio switching and its guards, destination and frequency-slot resolution, and the reception path.

## What is new since these were drafted in the firmware tree

- **Frequency slots.** `offer_frequency_slot` on the air, and a slot on the offer and on every destination. The client page states the derivation rule, since the field is omitted whenever a receiver can work the slot out for itself.
- **`ChannelIdentity`.** The offered channel is now name and PSK rather than a full `ChannelSettings`. The tag numbers match, so the change is wire compatible in both directions and no client needs updating to keep decoding beacons — that is stated explicitly, because the natural assumption on seeing a type change is the opposite.
- **Two ways to name a destination.** By value for a remote administrator, who cannot enumerate the channel table; by index for a locally connected client, which can. Mutually exclusive, with the indexed list winning if both arrive.
- **By-value channels are upserted** into the channel table on write, never evicting a live slot and never taking the primary. A channel named by value is in the table or it is not used: if no slot is free the field is withheld and its whole group cleared, for the offer as well as the target.
- **A 233-byte ceiling on remote writes**, refused with `Routing.Error TOO_LARGE` rather than half-stored, because the read-back would truncate to an empty payload and read as success.

## Corrections to previously documented behaviour

Three things the firmware-tree drafts described are no longer true, and one of them was actively misleading:

- A `channel_index` that is out of range or points at a disabled slot was documented as **falling back to the node's primary channel**. It is now skipped — no beacon is sent for that target. A client warning users about the old behaviour would be warning about the wrong risk.
- `broadcast_send_as_node` had its own section and a sanitisation row. The field is `reserved 3` as of meshtastic/protobufs#1047; both are gone.
- An invalid preset was documented as clearing the associated channel too. Only a value that is not a preset at all is cleared now; one the region cannot run is kept, because it describes a mesh elsewhere.

## Firmware these pages describe

meshtastic/firmware#11662 as it now stands, which includes three defects found after the pages were first drafted and fixed since:

- A preset-only target whose preset belongs to an EU sibling region never transmitted. The swap was reverted at key-up and the packet dropped for a preset the node's own region cannot run - the `NARROW_SLOW` case these pages use as their headline example.
- An offer consisting only of a pinned frequency slot was never sent, although the listener counted one as offer content. Sender and listener disagreed about what an offer is.
- Channel identity matched on raw PSK bytes, so a table holding a 1-byte shorthand and an offer naming the expanded key were treated as different channels and a duplicate slot was placed.

The pages describe the behaviour after those fixes, not before.

## Verification

- `pnpm lint:mdx` passes, with no warnings on any of the three pages.
- Every protocol claim was checked against the `.proto` definitions and the firmware source rather than carried over from the original documents.

The full production build has not been run locally, so CI is the first real check of cross-page links.


NomDeTom:
Superseded by #2644, opened from a branch in this repo so the Vercel preview can build without a team authorization step. Same commits, same content.

PR #846, opened 2023-10-10
Versioning Prep: Convert absolute to relative links

pdxlocations:
This PR prepares the site for versioning by changing all absolute file paths to relative paths.
[preview link](https://sandbox-git-relative-links-01-pdxlocations.vercel.app/)

Follows the [Recommended Practices from Docusaurus documentation on versioning](https://docusaurus.io/docs/versioning#recommended-practices)

**Except:**
`Refer to other docs by relative file paths with the .md extension, so that Docusaurus can rewrite them to actual URL paths during building. Files will be linked to the correct corresponding version.`
```md
The [@hello](hello.mdx#paginate) document is great!

See the [Tutorial](../getting-started/tutorial.mdx) for more info.
```
This practice requires manually editing every link on the site and I'm not clear on the benefits of this.

PR #2648, opened 2026-09-04
Enable Docusaurus versioning: 2.8 at /docs/, freeze 2.7 and below

rcarteraz:
## What this does

Enables Docusaurus docs versioning. The live docs become **2.8** at `/docs/`, and the pre-2.8 state is frozen and kept browsable at `/docs/2.7/`.

Per the discussion in #docs: tag the current documentation, freeze it, and open 2.8 from the same base. Right now a reader on shipped firmware sees documentation for features their radio does not have, and reports them as broken.

## Why 2.8 sits at the root

`lastVersion: "current"` keeps 2.8 on the unprefixed path, so **every existing `/docs/…` URL stays byte-identical**. That is what makes this cheap rather than disruptive:

- the ~51 redirect rules in `vercel.json` keep working, unedited
- the Algolia index is untouched — current pages still emit `docusaurus_tag: docs-default-current`, so site search behaves exactly as today
- Crowdin is untouched: its source glob is `docs/**/*.mdx` targeting `…/current/…`, and `versioned_docs/` sits outside it
- the root-level slugs (`/docs/introduction`, `/docs/faq`, `/docs/developers`) and the e2e specs need no changes

The alternative shape — 2.7 at `/docs/` with 2.8 at `/docs/next/` — would have moved every published URL and required converting 335 absolute links across the live tree first. That was attempted in #846 and abandoned.

## What's in it

**Archive presentation.** 2.7 is labelled "2.7 and below" so the snapshot reads as covering every pre-2.8 release, and carries Docusaurus's built-in `unmaintained` banner linking to the current page. `editUrl` is now a function returning `undefined` for frozen versions, so archived pages show no "Edit this page" link.

**Kept out of search engines.** Excluded from the sitemap via `ignorePatterns`, and `Disallow: /docs/2.7/` in `robots.txt`. Google should keep sending everyone to current docs.

**`scripts/freeze-version-links.mjs`** — run once per cut, right after `docs:version`. This is the part worth reviewing closely. `docs:version` copies `docs/` verbatim, so a snapshot inherits every root-absolute `/docs/…` link, and those still resolve — to the *live* version. An archived page silently walks readers into current docs, and `onBrokenLinks` cannot catch it because the route genuinely exists. The script repoints seven link forms at the frozen version, converts file-path links (`/docs/foo/bar.mdx`) to relative form so `slug:` overrides resolve correctly, is idempotent, and **fails the run** if anything still escapes the version.

**Read-only enforcement.** A `frozen-docs` CI job plus CODEOWNERS entries. The job is scoped to `modified|deleted|renamed`, so cutting a new version (all additions) passes while editing an existing snapshot fails; a `version-cut` label overrides it for deliberate corrections.

**One real bug fixed.** The custom mobile sidebar (`src/theme/Navbar/MobileSidebar/PrimaryMenu/index.tsx`) renders unrecognised navbar item types as a bare link, so the version dropdown would have shipped as `href="undefined"` — invisible during desktop review.

The locale dropdown is removed from the navbar for now. `i18n` config, `crowdin.yml` and the `i18n/en/` sources are all untouched; this only stops advertising it while translation is unresolved.

## Reviewing this

Commit 1 is 359 files, but only ~10 are hand-written — the rest is the mechanical `versioned_docs/version-2.7/` snapshot. Worth actual attention: `docusaurus.config.js`, `scripts/freeze-version-links.mjs`, the `frozen-docs` CI job.

Commits 2 and 3 are separate and small:

- **Heltec LoRa32 V4-R8 docs**, content taken from #2581 (still a draft there). Its 1.33 MB PNG pinmap is converted to WebP, 243 KB.
- **Two broken image references** that were 404ing silently, since Docusaurus does not verify static asset paths: `rak5005-0_connectors.web` was missing its trailing `p`, and the BaseUI clock screenshot pointed at a file that was never committed — only green/red variants were, so it now uses the `Light`/`Dark` pair like every other screenshot on that page.

#2338 was also considered and deliberately **not** taken: its T-Beam 1W content is already on master in better form via #2553, and it would have introduced a broken image reference and a wrong device image.

## Verification

- `pnpm run build` passes with `onBrokenLinks: "throw"`; 337 routes per version
- `pnpm test:e2e` — 42/42
- `pnpm lint:mdx` passes
- zero absolute links escape `/docs/2.7/` in archived page bodies, checked in the built HTML
- zero missing image references across the whole built site (was 2 before this branch)
- sitemap contains 409 URLs, none from `/docs/2.7/`

Note for anyone running the e2e suite locally: `reuseExistingServer` is enabled outside CI, so if you have another Docusaurus dev server on port 3000 the suite will silently test *that* site instead of this build.

## Follow-ups, not in this PR

- The "Documentation Organization" table in `docs/development/documentation/index.mdx` still refers to `docs/settings` and `docs/developers`, renamed some time ago to `docs/configuration` and `docs/development`. Left alone pending the separate docs-standards work.
- `docs/` still carries 335 root-absolute `/docs/…` links. They are correct today, but converting them to relative is what would make each future cut nearly free.
- At 3.0: `pnpm docusaurus docs:version 2.8 && node scripts/freeze-version-links.mjs 2.8`, then add a `versions` entry. Note the config's version options must be temporarily removed to re-cut an *existing* version, since Docusaurus validates them against `versions.json` before the command runs.


caveman99 (approved):

pdxlocations (approved):
Built locally and behaves as expected.  Nice!

PR #2608, opened 2026-08-19
docs: add AS3935 lightning sensor to telemetry table

ndoo:
## What did you change
Added the AS3935 lightning sensor to the "Currently Supported Sensor Types" table, with its I2C addresses and reported data points, plus a note that `AS3935_IRQ` avoids polling the sensor over I2C every second.

Fixed column padding on the MAX30102/MLX90614/MLX90632 rows to match the rest of the table.

## Why did you change it
meshtastic/firmware#10931 added AS3935 lightning-detection sensor support to the Environment Telemetry module.

## Screenshots
None provided, change is small.

## Summary by CodeRabbit

* **Documentation**
  * Added AS3935 to the supported sensor list.
  * Documented its I²C addresses and lightning telemetry fields.
  * Added wiring guidance for the IRQ pin to reduce unnecessary polling.

PR #2659, opened 2026-09-08
Add the sensors missing from the Telemetry Module table

ndoo:
## What did you change

Added the sensors that Meshtastic firmware [2.8.0 alpha (`v2.8.0.47db0e3`)](https://github.com/meshtastic/firmware/releases/tag/v2.8.0.47db0e3) auto-detects but the Telemetry Module page's "Currently Supported Sensor Types" table was missing:

- **ADS1X15** - ADC channel voltages (4 channels per device, 8 with a second device)
- **AS3935** - lightning strike count and estimated storm distance
- **BH1750** - light intensity
- **DS18B20** (via a DS2482 1-Wire bridge) - temperature, up to 8 probes
- **HM330X** - particulate matter concentration by size
- **MAX17048** - battery voltage and charge level
- **PCT2075** - temperature
- **SCD30** - CO₂, temperature, humidity
- **SCD4X** - CO₂, temperature, humidity
- **SEN5X** - particulate matter; temperature/humidity and VOC/NOx index depend on the model (SEN50/54/55)
- **SEN6X** - particulate matter and temperature/humidity; VOC/NOx index, CO₂ and formaldehyde depend on the model (SEN62/63C/65/66/68/69C)
- **SFA30** - formaldehyde, temperature, humidity
- **SPA06** - barometric pressure, temperature
- **TSL2561** - light intensity

Each new row lists the sensor's I²C address(es) and the data points it reports, plus a note about wiring the AS3935 IRQ pin so it isn't polled over I²C every second. The SEN5X and SEN6X rows carry a per-model capability matrix in a footnote, since the reported metrics differ between Sensirion variants.

The new rows are appended at the end of the existing table and sorted among themselves, so no existing row moves or changes wording. The first commit is those additions as insertions only, so the new content is easy to review; the second commit is an `oxfmt` whitespace-only reflow of the table's column padding.

Only `docs/configuration/module/telemetry.mdx` changes. The frozen `versioned_docs/version-2.7` copy is deliberately left alone, since `docs/` now tracks 2.8. Built locally with `pnpm build`; `pnpm lint:mdx` and `oxfmt --check` pass.

## Why did you change it

The table had drifted well behind the firmware, missing roughly a dozen sensors added across the 2.7.x series and for 2.8.0, with SCD4x / CO₂ support previously only documented on the Android app pages. The list here was read from that release's source: the drivers under `src/modules/Telemetry/Sensor/`, the I²C address defines in `src/configuration.h`, the probing logic in `src/detect/ScanI2CTwoWire.cpp`, and which telemetry module reads each sensor.

This supersedes #2608, which added AS3935 alone; its AS3935 IRQ note is carried over here. It also overlaps #2253 on the SEN5X / SCD / SFA30 rows, whose other additions (an air-quality screen-enabled option and an I²C-clock limitations section) are out of scope here.

### Before

pr-before-table

### After

pr-after-table

pr-after-footnotes

## Summary by CodeRabbit

- **Documentation**
  - Expanded the Telemetry Module’s supported sensor list, including ADS1X15, BH1750, DS18B20, HM330X, MAX17048, PCT2075, SCD30, SCD4X, SEN5X, SEN6X, SFA30, SPA06, and TSL2561.
  - Improved table formatting and clarified sensor-specific details, including DS18B20 probe limits and SPA06 compatibility.
  - Added inline guidance for AS3935 IRQ wiring and documented the capabilities of SEN5X and SEN6X models.


docs/configuration/module/telemetry.mdx
```diff
@@ -29,39 +29,57 @@ Supported sensors connected to the I2C bus of the device will be automatically d
...
+|     DFROBOT_RAIN     |          0x1d          |                                        Tip Bucket Rain counter                                        |
+|       RadSens        |          0x66          |                                            Radio Dosimeter                                            |
+|       MAX30102       |          0x57          |                          Heart Rate, Oxygen Saturation, and body temperature                          |
+|       MLX90614       |          0x5A          |                                           Body temperature                                            |
+|       MLX90632       |          0x3A          |                                           Body temperature                                            |
+|       NAU7802        |          0x2A          |                             24-Bit differential ADC for Wheatstone bridge                             |
+|       ADS1X15        | 0x48, 0x49, 0x4A, 0x4B |                                ADC channel voltages (up to 8 channels)                                |
```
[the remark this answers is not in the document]
ndoo: Fixed in 74ad9c2b. The cell now reads "ADC channel voltages (4 channels per device, 8 with two devices)" — `ADS1X15Sensor.cpp` reads four single-ended channels per device, and eight is only reached with a second device (`ADS1X15` at 0x48 → ch0-3, `ADS1X15_ALT` at 0x49-0x4B → ch4-7).

docs/configuration/module/telemetry.mdx
```diff
@@ -29,39 +29,57 @@ Supported sensors connected to the I2C bus of the device will be automatically d
...
+|        HM330X        |          0x40          |                                      Concentration units by size                                      |
+|       MAX17048       |          0x36          |                                   Battery voltage and charge level                                    |
+|       PCT2075        |          0x37          |                                              Temperature                                              |
+|        SCD30         |          0x61          |                               CO<sub>2</sub>, temperature and humidity                                |
+|        SCD4X         |          0x62          |                               CO<sub>2</sub>, temperature and humidity                                |
+|        SEN5X         |          0x69          |                Particulate matter by size, VOC and NOx index, temperature and humidity                |
+|        SEN6X         |          0x6B          | Particulate matter by size, CO<sub>2</sub>, VOC and NOx index, temperature, humidity and formaldehyde |
```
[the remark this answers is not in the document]
ndoo: Fixed in 74ad9c2b. Both rows now read "particulate matter by size; further metrics vary by model" and carry a per-model capability matrix in a footnote (temperature/humidity, VOC index, NOx index, CO₂, formaldehyde per variant), taken from `SENXXSensor.cpp::updateCapabilities()`.

ndoo:
Screenshots updated with the footnote migration

docs/configuration/module/telemetry.mdx
```diff
@@ -29,39 +29,74 @@ Supported sensors connected to the I2C bus of the device will be automatically d
...
+|     DFROBOT_LARK     |          0x42          | Temperature, barometric pressure, humidity, wind direction, wind speed |
+|     DFROBOT_RAIN     |          0x1d          |                        Tip Bucket Rain counter                         |
+|       RadSens        |          0x66          |                            Radio Dosimeter                             |
+|       MAX30102       |          0x57          |          Heart Rate, Oxygen Saturation, and body temperature           |
+|       MLX90614       |          0x5A          |                            Body temperature                            |
+|       MLX90632       |          0x3A          |                            Body temperature                            |
+|       NAU7802        |          0x2A          |             24-Bit differential ADC for Wheatstone bridge              |
+|       ADS1X15        | 0x48, 0x49, 0x4A, 0x4B |    ADC channel voltages (4 channels per device, 8 with two devices)    |
+|   AS3935[^as3935]    |    0x01, 0x02, 0x03    |        Lightning strike count (1h) and estimated storm distance        |
```
rcarteraz: "estimated storm distance" has no unit. The field is `lightning_distance_km`, so say km. "(1h)" would read better as "in the past hour".

docs/configuration/module/telemetry.mdx
```diff
@@ -29,39 +29,74 @@ Supported sensors connected to the I2C bus of the device will be automatically d
...
+|        BH1750        |          0x23          |                            Light intensity                             |
+| DS18B20 (via DS2482) |       0x18–0x1F        |                  Temperature (up to 8 1-Wire probes)                   |
```
rcarteraz: `DS248XSensor.cpp` also handles the DS2484, which this misses, and "up to 8 probes" holds only on the DS2482-800. The DS2484 is single channel.

docs/configuration/module/telemetry.mdx
```diff
@@ -29,39 +29,74 @@ Supported sensors connected to the I2C bus of the device will be automatically d
...
+|        HM330X        |          0x40          |                      Concentration units by size                       |
+|       MAX17048       |          0x36          |                    Battery voltage and charge level                    |
+|       PCT2075        |          0x37          |                              Temperature                               |
+|        SCD30         |          0x61          |                CO<sub>2</sub>, temperature and humidity                |
+|        SCD4X         |          0x62          |                CO<sub>2</sub>, temperature and humidity                |
```
rcarteraz: Inline HTML in translated prose gets corrupted on Crowdin round trips, and `lint-mdx.sh` does not check `sub`. Plain CO2 reads the same aloud and matches a find-in-page for "CO2", which the Unicode subscript does not.

docs/configuration/module/telemetry.mdx
```diff
@@ -29,39 +29,74 @@ Supported sensors connected to the I2C bus of the device will be automatically d
...
+|        SEN5X         |          0x69          |   Particulate matter by size; further metrics vary by model [^sen5x]   |
+|        SEN6X         |          0x6B          |   Particulate matter by size; further metrics vary by model [^sen6x]   |
+|        SFA30         |          0x5D          |                 Formaldehyde, temperature and humidity                 |
+|        SPA06         |       0x76, 0x77       |                    Barometric pressure, Temperature                    |
```
rcarteraz: This row promises more than the build delivers. `EnvironmentTelemetry.cpp:287` is `#if HAS_SPA06 && __has_include(<Adafruit_SPA06_003.h>)`, and `HAS_SPA06` is defined only in `variants/nrf52840/seeed_mesh_tracker_X1/variant.h`. Everywhere else the scanner identifies the chip and no sensor is registered. Either qualify the row or drop it.

docs/configuration/module/telemetry.mdx
```diff
@@ -29,39 +29,74 @@ Supported sensors connected to the I2C bus of the device will be automatically d
...
+|       TSL2561        |          0x29          |                            Light intensity                             |
+
+[^as3935]: Wiring the IRQ pin and defining `AS3935_IRQ` in the variant avoids polling the sensor over I2C every second.
```
rcarteraz: [Section 11.2](https://github.com/meshtastic/design/blob/master/standards/meshtastic_design_standards_v1_5.md#112-voice) wants the imperative for the step rather than a gerund: "Wire the IRQ pin and define `AS3935_IRQ` in the variant to avoid an I&sup2;C read every second." That also settles I2C against the I&sup2;C the table header uses.

docs/configuration/module/telemetry.mdx
```diff
@@ -29,39 +29,74 @@ Supported sensors connected to the I2C bus of the device will be automatically d
...
+
+[^sen5x]: All SEN5X models report particulate matter by size.
+
+    | Model | Temperature and humidity | VOC index | NOx index |
+    | :---: | :----------------------: | :-------: | :-------: |
+    | SEN50 |         &mdash;          |  &mdash;  |  &mdash;  |
+    | SEN54 |         &check;          |  &check;  |  &mdash;  |
+    | SEN55 |         &check;          |  &check;  |  &check;  |
```
rcarteraz: These matrices would read better as a section below the table than as a footnote table. Two reasons beyond the extra table: an em dash at default screen reader punctuation settings is not spoken, so a cell meaning "no" comes through as silence, and GFM emits the model names as `td` with no `scope="row"`, so reading the matrix cell by cell never says which model the row belongs to.

Suggested replacement, below the table:

```markdown
#### SEN5X models

All SEN5X models report particulate matter by size. The SEN54 and SEN55 also report temperature, humidity and VOC index. The SEN55 adds NOx index.
```

The table cell can then point down to the section instead of carrying a footnote marker.

docs/configuration/module/telemetry.mdx
```diff
@@ -29,39 +29,74 @@ Supported sensors connected to the I2C bus of the device will be automatically d
...
+[^sen6x]: All SEN6X models report particulate matter by size.
+
+    | Model  | Temperature and humidity | VOC and NOx index | CO<sub>2</sub> | Formaldehyde |
+    | :----: | :----------------------: | :---------------: | :------------: | :----------: |
+    | SEN62  |         &check;          |      &mdash;      |    &mdash;     |   &mdash;    |
+    | SEN63C |         &check;          |      &mdash;      |    &check;     |   &mdash;    |
+    | SEN65  |         &check;          |      &check;      |    &mdash;     |   &mdash;    |
+    | SEN66  |         &check;          |      &check;      |    &check;     |   &mdash;    |
+    | SEN68  |         &check;          |      &check;      |    &mdash;     |   &check;    |
+    | SEN69C |         &check;          |      &check;      |    &check;     |   &check;    |
```
rcarteraz: Same here.

```markdown
#### SEN6X models

All SEN6X models report particulate matter by size, temperature and humidity. The SEN65, SEN66, SEN68 and SEN69C add VOC and NOx index. The SEN63C, SEN66 and SEN69C report CO2. The SEN68 and SEN69C report formaldehyde.
```

This also removes the third `CO<sub>2</sub>`, in the header on line 92. Grouping by metric keeps it shorter than six model rows, but say if you would rather it read model by model.

rcarteraz (changes requested):
CI is green and the admonition is gone, thanks. A few things inline, mostly small.

The section numbers refer to [section 11 of the Meshtastic client design standards](https://github.com/meshtastic/design/blob/master/standards/meshtastic_design_standards_v1_5.md#11-documentation), which covers documentation and landed in meshtastic/design#150 this week. It governs new and edited writing rather than a sweep of what is already on the page, so nothing here is about the rows that were already in the table.

One point that does not fit on a line: ADC, VOC, NOx and IRQ are not expanded anywhere on the page, and [section 11.3](https://github.com/meshtastic/design/blob/master/standards/meshtastic_design_standards_v1_5.md#113-plain-language) asks for an expansion on first use because readers arrive from search. ADC, VOC and NOx are also worth adding to the [glossary](https://meshtastic.org/docs/terms/), where none of the three appear today and where they would help beyond this page. Suggested entries, in the definition list shape the glossary already uses:

```markdown
ADC
: Analog-to-digital converter. A circuit that converts an analog voltage into a numeric reading, used for battery voltage and for external analog inputs on sensors such as the ADS1X15.

NOx
: Nitrogen oxides. Some air quality sensors report these as an index rather than a concentration.

VOC
: Volatile organic compounds. Some air quality sensors report these as an index rather than a concentration.
```

IRQ is specific to the AS3935 wiring note, so expanding it on the page is enough.

Everything else checked out against v2.8.0.47db0e3: the addresses, the data points, and both Sensirion capability sets match the drivers.

rcarteraz:
Following up on the glossary point in my review above: #2678 is open and touches the same ground, so it is worth reading the two together.

That PR replaces the definition list at `/docs/terms/` with a generated page backed by `glossary/glossary.json`, so the entries I suggested land in a different shape — JSON objects rather than `term` / `: definition` pairs. All three are now in it: ADC, VOC and NOx.

It also adds hover tooltips for terms in prose, and they work inside tables. Once both land, ADC, VOC and NOx will be defined on hover directly in the sensor rows this PR adds, which covers the section 11.3 concern without expanding each one inline.

Nothing here blocks this PR, and there is no conflict — #2678 does not touch `telemetry.mdx`. Whichever merges second just picks up the other's behaviour.

ndoo:
Thanks for the notes. I’ll give rewriting the noted sections a go later
this weekend.

One question though - if the SEN5x/SEN6x move out of footnotes, then shall
the AS3935 also move to below the sensor table?

ndoo:
@coderabbitai review

ndoo:
@rcarteraz Pushed changes that should address the review points, plus an oxfmt pass on the table.

docs/configuration/module/telemetry.mdx
ndoo: Fixed: now says "in the past hour" and states the unit (km).

docs/configuration/module/telemetry.mdx
ndoo: Fixed: row now covers both DS2482-800 (up to 8 probes) and DS2484 (1 probe).

docs/configuration/module/telemetry.mdx
ndoo: Fixed: dropped the <sub> tags, both rows now read plain "CO2".

docs/configuration/module/telemetry.mdx
ndoo: Qualified rather than dropped: row now reads "SPA06 (Mesh-Tracker-X1 only)".

docs/configuration/module/telemetry.mdx
ndoo: Fixed: imperative mood, IRQ expanded on first use, and I&sup2;C to match the header — moved into a `#### AS3935` section below the table.

docs/configuration/module/telemetry.mdx
ndoo: Moved into a `#### SEN5X models` section below the table, written per model rather than a table.

docs/configuration/module/telemetry.mdx
ndoo: Moved into a `#### SEN6X models` section below the table, per model, which also drops the third CO2 subscript along with it.

ndoo:
The screenshots are not updated, please use the Vercel workflow 😅 it’s quite a bit of churn to be updating the screenshot each review

rcarteraz (approved):

PR #2540, opened 2026-07-19
docs: normalize product/tech name casing (nRF52840, ESP32-S3, LoRa, WiFi, Node-RED)

bruschill:
## What did you change

A consistency-only terminology/casing sweep across the docs, normalizing product and technology names toward their official forms. **No prose or meaning changes** — only capitalization/spelling of names.

- `NRF52840` → `nRF52840` (10) — Nordic's official casing, in device spec tables/bullets.
- `ESP32S3` → `ESP32-S3` (4) — Espressif's official casing; matches the docs' dominant `ESP32-S3` usage.
- `Wi-Fi` → `WiFi` (13) — matches the docs' overwhelmingly dominant house style (`WiFi` ~200 vs `Wi-Fi` 13).
- `Lora` → `LoRa` (5) — **only** genuine radio-technology references ("LoRa front-end circuit", "wideband LoRa antenna", "915Mhz LoRa PCB Antenna", "Meshtastic LoRa Chirp").
- `Node-red` / `node-red` → `Node-RED` (2).

## Why did you change it

The docs mixed several casings for the same names, flagged in an accessibility/consistency review. These are the unambiguous, official-form corrections.

Each change was scoped to **prose only** and deliberately avoids clobbering:
- the Elecrow wiki **URL** that contains `NRF52840`,
- **LILYGO product/board names** — `T-Lora Pager`, `TTGO Lora`, `Lora V2.1`, etc. (LILYGO's own branding, left exactly as-is),
- `gateway-wifi` **JSX values** and `wifi_ssid`/`wifi_psk` config identifiers,
- lowercase `nrf52840` **tab values / CLI args**,
- image **file paths**.

**Intentionally left for maintainers:** the `meshtasticd` / `MeshtasticD` casing. The daemon's binary is lowercase, but the project brands its own landing page `title: MeshtasticD` and the name appears inside image file paths, so normalizing it is a judgment call I didn't want to make unilaterally.

Third of three separate docs PRs from the review (after component a11y and docs alt-text/links/headings).

## Screenshots

N/A — text-only casing changes with no visual or layout impact. Diff is a symmetric +34/−34 across 22 files.

## Summary by CodeRabbit

* **Documentation**
  * Standardized terminology and capitalization for WiFi, LoRa, nRF52840, ESP32-S3, and Node-RED.
  * Corrected hardware specifications and device table entries for improved accuracy.
  * Improved formatting and consistency across encryption, MQTT, Apple, hardware, and glossary documentation.
[...]

PR #2667, opened 2026-09-09
docs: standardize Wi-Fi and add a name-casing check

### LINKED ISSUE #2687: Standardize Wi-Fi in prose and check name casing in review
The docs spell the wireless standard three ways. Prose should use Wi-Fi with the dash everywhere, and a check should keep it that way.

Wi-Fi with the dash is what the Wi-Fi Alliance uses and what the Android client's word list already uses, so adopting it here makes the docs and the client guides agree. The earlier casing pass (meshtastic/meshtastic#2540) standardized nRF52840, ESP32-S3, LoRa, LoRaWAN and Node-RED but picked WiFi for this one, so this is the remaining name to settle.

Two parts:

1. Replace WiFi with Wi-Fi in prose across the current docs. Code, config keys, protobuf field names, file paths, URLs and verbatim tool output stay as they are, so `wifi_ssid`, vendor resource URLs and the esptool `Features: WiFi, BT` line are not touched. Frozen snapshots under `versioned_docs/` stay untouched. The synced Android and Apple pages are authored in the client repositories and would be reverted by the next sync, so they need separate PRs there.

2. Add a name-casing check to `docs-review.yml` covering all six names. It should read added lines only, and blank fenced blocks, inline code, URLs, link targets and HTML tags before matching, so it comments on prose rather than on identifiers. Non-blocking, posting a suggestion like the workflow's other checks.

Companion to meshtastic/design#150, which writes the casing rules into section 11 of the client design standards.

rcarteraz:
## What did you change

Wi-Fi now takes the dash across the docs, replacing 213 prose occurrences of WiFi in 62 files.

Left alone on purpose: code, config keys, protobuf field names, file paths, URLs and verbatim tool output. So `wifi_ssid` stays, the Heltec and Elecrow resource URLs stay, and the esptool `Features: WiFi, BT` output line stays. Frozen snapshots under `versioned_docs/` are untouched, and the synced Android and Apple pages are excluded because they are authored in the client repositories and would be reverted by the next sync. Those 18 occurrences need PRs in the client repos.

Also adds a casing check to `docs-review.yml`, covering the normalizations from #2540 as well as Wi-Fi: LoRa, LoRaWAN, nRF52840, ESP32-S3, Node-RED and meshtasticd. It reads added lines only, and blanks fenced blocks, inline code, URLs, link targets and HTML tags before matching, so it comments on prose rather than on identifiers. It posts a suggestion like the workflow's other checks and does not block.

## Why did you change it

#2540 standardized five product names and picked WiFi for this one. Wi-Fi with the dash is what the Wi-Fi Alliance uses and what the Android client's word list already uses, so taking the dash here makes the docs and the client guides agree without anyone reconciling them later. This is the one-off to get there; the check keeps it from drifting back.

The other four names from #2540 are already at zero violations, so this PR does not change them. The check protects all five going forward.

Companion to meshtastic/design#150, which writes the casing rules into section 11 of the client design standards.

Closes #2687

## Summary by CodeRabbit

- **Documentation**
  - Standardized “Wi-Fi” terminology across product guides, hardware specifications, glossary entries, blog content, and integration documentation.
  - Corrected a minor “gyroscope” spelling error in hardware documentation.

- **Quality Improvements**
  - Documentation reviews now check newly added content for incorrect product and technology name casing.
  - Review checks now focus on added lines and exclude code, URLs, links, HTML, and fenced code blocks when evaluating prose and import usage.


rcarteraz:
@bruschill Was speaking with @jamesarich and given the Apple and Android docs were predominantly using the correct form, Wi-Fi, we opted to fix it in the docs (but only for the latest version). Which is why I opened the PR. I also added it to the docs-review.yml

bruschill:
> @bruschill Was speaking with @jamesarich and given the Apple and Android docs were predominantly using the correct form, Wi-Fi, we opted to fix it in the docs (but only for the latest version). Which is why I opened the PR. I also added it to the docs-review.yml

I appreciate this!

.github/workflows/docs-review.yml
```diff
@@ -120,11 +123,80 @@ jobs:
                 }
               }
 
+              // Check product and technology name casing in prose.
+              // Skips fenced blocks, inline code, URLs and link targets, since
+              // those are literal strings rather than prose.
+              if (/\.mdx?$/.test(filename) && patch) {
+                const CASING = [
+                  [/\bWiFi\b/g, "Wi-Fi"],
```
bruschill: Design standard 11.4 rejects lora, esp32s3, wifi, and WIFI, but the regex table does not match them.

```
[/\b(?:WiFi|wifi|WIFI)\b/g, "Wi-Fi"],
[/\b(?:Lora|LORA|lora)\b/g, "LoRa"],
[/\b(?:ESP32S3|esp32s3)\b/g, "ESP32-S3"],
...
```

.github/workflows/docs-review.yml
```diff
@@ -120,11 +123,80 @@ jobs:
...
+                  [/\bLorawan\b|\bLoraWAN\b/g, "LoRaWAN"],
+                  [/\bNRF52840\b/g, "nRF52840"],
+                  [/\bnrf52840\b/g, "nRF52840"],
+                  [/\bESP32S3\b/g, "ESP32-S3"],
+                  [/\bNode-red\b|\bnode-red\b/g, "Node-RED"],
+                  [/\bMeshtasticd\b|\bMeshtasticD\b/g, "meshtasticd"],
+                ];
+
+                const found = new Map();
+                let inFence = false;
```
bruschill: `inFence` persists across `@@` hunk gaps. An omitted closing fence can cause later prose additions to be skipped. A hunk starting inside an existing fence treats code as prose. Inspect the full head file and use the patch only to identify added lines.

```js
   function addedLineNumbers(patch) {
     const added = new Set();
     let newLine = null;

     for (const raw of patch.split("\n")) {
       const hunk = raw.match(
         /^@@ -\d+(?:,\d+)? \+(\d+)(?:,\d+)? @@/
       );

       if (hunk) {
         newLine = Number(hunk[1]);
       } else if (newLine === null || raw.startsWith("\\ No newline")) {
         continue;
       } else if (raw.startsWith("+")) {
         added.add(newLine++);
       } else if (raw.startsWith(" ")) {
         newLine++;
       }
       // Removed lines do not advance the new-file line number.
     }

     return added;
   }

   function fenceMarker(line) {
     return line.match(/^ {0,3}(`{3,}|~{3,})(.*)$/);
   }

   const addedLines = addedLineNumbers(patch);

   const { data: blob } = await github.rest.git.getBlob({
     owner,
     repo,
     file_sha: file.sha,
   });

   const contents = Buffer.from(
     blob.content.replace(/\n/g, ""),
     "base64"
   )
     .toString("utf8")
     .replace(/\r\n/g, "\n");

   let fence = null;

   for (const [index, line] of contents.split("\n").entries()) {
     const marker = fenceMarker(line);

     if (marker) {
       const run = marker[1];
       const trailing = marker[2];

       if (!fence) {
         // Backtick fence info strings cannot contain backticks.
         if (run[0] !== "`" || !trailing.includes("`")) {
           fence = { character: run[0], length: run.length };
         }
       } else if (
         run[0] === fence.character &&
         run.length >= fence.length &&
         /^[ \t]*$/.test(trailing)
       ) {
         fence = null;
       }

       continue;
     }

     const lineNumber = index + 1;
     if (fence || !addedLines.has(lineNumber)) continue;

     // Apply the existing inline-code, URL, link, and HTML masking here.
     scanProse(line);
   }
 ```

File: .github/workflows/docs-review.yml [hunk 3]
```diff
@@ -120,11 +123,80 @@ jobs:
...
+                  if (!isAdded && !isContext) continue;
+                  const line = raw.slice(1);
+                  if (/^\s*(```|~~~)/.test(line)) {
+                    inFence = !inFence;
+                    continue;
+                  }
+                  if (!isAdded || inFence) continue;
+
+                  // Blank out the parts that are not prose, keeping offsets stable.
+                  const prose = line
```
bruschill: The masking misses double-backtick code spans, reference link targets such as [driver]: ./WiFi.md, multiline JSX tags, and import paths. These produce false casing suggestions. 

Maybe an MDX parser would be better here?

.github/workflows/docs-review.yml
bruschill: `pulls.listFiles` returns one page. Checks silently skip files after 100. Use github.paginate(github.rest.pulls.listFiles, ...) instead.

I hope we never run into this w/ 100+ changed files

bruschill (commented):
Some small requests, nothing major!

.github/workflows/docs-review.yml
rcarteraz: Right, and the gap was wider than the four you named — the 11.4 table also lists `bluetooth`, `mqtt`, `Mqtt`, `gps`, `Gps` and `BT`. All in now except `BT`, with guards so the lowercase forms do not fire on identifiers.

I measured before and after by running the workflow script itself over all 325 doc files with every line treated as added. The naive lowercase additions flagged 70 files, and most were not real: front matter `id: bluetooth`, dotted config keys like `lora.region` and `mqtt.enabled`, and path segments like `radio/lora`. With front matter skipped and a guard that rejects a term inside a dotted key or a path but still allows one ending a sentence, it is 27 files, and the ones I spot-checked are genuine — "not functional on MeshtasticD devices", "transport to clients, mqtt, and UPD".

`BT` is the one place I did not follow the table. It is the column header in the hardware tables on 21 pages, `| Name | MCU | Radio | Wi-Fi | BT | GPS |`, where Bluetooth does not fit the column, so flagging it means a suggestion on every hardware page edit. Happy to add it if you would rather the check match the table exactly and the tables get rewritten.

.github/workflows/docs-review.yml
rcarteraz: Adopted, essentially as you wrote it. The patch now only works out which line numbers the change adds, and fence state comes from the head file via `getBlob`, including your fence matching rules: run length and character have to match to close, and a backtick info string cannot contain a backtick.

You were right that this was not a tuning problem. I had already patched the same area once by consuming context lines for fence tracking, which narrowed it but could not fix it — a hunk that begins inside a fence has no opening marker anywhere in the diff, so no amount of patch reading reaches it.

Two cases from the unit tests, both of which the old code got wrong and both passing now: an added line inside a pre-existing fence is silent, and an added prose line after a fence the patch never shows is still flagged. I also tested that removed lines do not shift the new-file line number, which your `addedLineNumbers` handles and my previous line counting would not have.

.github/workflows/docs-review.yml
rcarteraz: Fixed the concrete ones. Code spans now match any backtick run, so ``double`` is masked; reference link definitions and `import`/`export` lines are skipped whole; multi-line tags carry an open-tag state across lines; and JSX expressions are skipped by brace depth, which is what had been flagging `{label: "Lora V3.0"}` in the LILYGO tab lists.

On the parser, you are right that regex masking has a ceiling. I would rather not pull remark or the MDX compiler into this workflow though: it is an advisory bot that posts suggestions and never blocks, so a miss is mild, while a parse step adds install time and a new way for the comment job to fail.

The better home for a parser-based check is `scripts/lint-mdx.sh`, which already runs in CI as `pnpm run lint:mdx`. That turns casing from a suggestion into a real check with a real AST, and the two would not duplicate each other. Feels like a follow-up issue rather than something to grow this PR into — say the word and I will open one.

.github/workflows/docs-review.yml
rcarteraz: Fixed — `github.paginate(github.rest.pulls.listFiles, ...)` now. Worth noting `generated-docs-guard.yml` already paginates and fails closed when the count does not match, so this workflow was the inconsistent one.

On never hitting 100 files: this PR is 63 and #2668 is 43, so it is closer than it sounds. A sweep like this one, or a sync PR, would have silently skipped the tail.

PR #2668, opened 2026-09-09
docs: use the @site alias for imports

rcarteraz:
## What did you change

Converted every relative import in `docs/` to the canonical Docusaurus `@site` alias. 45 imports across 42 files:

- `/src/components/DeviceImage` and the other `/src/...` paths become `@site/src/...` (42 imports)
- the three relative `_lora-regions.mdx` includes become `@site/docs/blocks/_lora-regions.mdx`, matching how `_qr-code.mdx` and the other blocks are already imported

No other changes. Nothing outside import lines was touched.

## Why did you change it

`docs-review.yml` flags relative import paths and asks for the `@site/src/` alias. Six files came up on #2667, but only because that PR happened to touch them — the same pattern was present in 42 files. Converting all of them now means the remaining ones don't get flagged one PR at a time as people edit unrelated things.

Verified with `docusaurus build`: server and client both compiled successfully, so every converted path resolves.

Note this overlaps #2667 on some of the same files, but on different lines — that PR changes prose, this one changes import statements. Whichever merges second may want a rebase.

## Summary by CodeRabbit

- **Documentation**
  - Updated documentation component references to use the site’s standardized path aliases.
  - Improved consistency across FAQ, configuration, getting-started, development, Meshtasticd, and hardware documentation pages.
  - Replaced the static link-budget image with an interactive chart in the radio settings documentation.
  - Updated formatting for improved readability.
  - No other user-facing documentation content or functionality was changed.

### DIFF
--- a/docs/about/faq.mdx
+++ b/docs/about/faq.mdx
@@ -6,8 +6,8 @@ sidebar_position: 3
 description: "Find answers to your Meshtastic questions: from getting started, device setup, to advanced features. Learn about app compatibility, channel sharing, and ham license benefits. Dive into our comprehensive FAQ."
 ---
 
-import { FaqAccordion } from "/src/components/FaqAccordion";
-import { FaqStructuredData } from "/src/components/FaqStructuredData";
+import { FaqAccordion } from "@site/src/components/FaqAccordion";
+import { FaqStructuredData } from "@site/src/components/FaqStructuredData";
 import QRCode from "@site/docs/blocks/_qr-code.mdx";
 
 export const Faq = {
--- a/docs/about/overview/radio-settings.mdx
+++ b/docs/about/overview/radio-settings.mdx
@@ -7,7 +7,7 @@ sidebar_position: 1
 description: "Maximize your Meshtastic device's potential with detailed radio settings instructions, including frequency bands, data rates, and encryption options."
 ---
 
-import { FrequencyCalculator } from "/src/components/tools/FrequencyCalculator";
+import { FrequencyCalculator } from "@site/src/components/tools/FrequencyCalculator";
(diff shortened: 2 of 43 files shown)


NomDeTom:
Clod looked at it:

The JSDoc usage examples in `src/components/DeviceImage.tsx` (lines 48, 98, 131, 221) still read

    import { DeviceImage } from "/src/components/DeviceImage";

That's the most-imported component in the repo, and those four lines are the source the bot will keep flagging as people paste them into new pages. Worth folding in here or picking up as a quick follow-up.

Separately, `versioned_docs/version-2.7/` still has 39 files with the old form. Probably deliberate for a frozen snapshot — flagging in case it isn't.

rcarteraz:
@NomDeTom good catch on `DeviceImage.tsx` — folded in. All four JSDoc examples now read `@site/src/components/DeviceImage`. That was the gap that mattered: converting the pages while leaving the examples people paste from would have kept reintroducing the pattern, and it is the most imported component in the repo.

`versioned_docs/version-2.7/` is left alone deliberately. The snapshot is frozen and CODEOWNERS-protected, so it is not a PR target.

Also rebased onto master, since #2669 and #2671 landed and `radio-settings.mdx` conflicted. Kept both imports there, with `FrequencyCalculator` on the alias alongside the new `LinkBudgetChart`. `docusaurus build` passes, so the resolution and the conversions all resolve.

Worth noting the check itself would not have caught the JSDoc lines, since it only looks at `.mdx`. Extending it to `.tsx` would mostly flag real component imports, which are supposed to be relative, so the example lines are probably better handled by review than by a rule.

NomDeTom (approved):

PR #2674, opened 2026-09-11
Docs/base url

NomDeTom:
## What did you change

Raw HTML in MDX keeps its paths exactly as written. Docusaurus rewrites Markdown image and link paths against `baseUrl`, but it never looks inside an `` tag, so fifty-one references across twenty-one files point at paths like `/img/hardware/screen.webp` and resolve against the server root.

- **20 pages** in `docs/` and `blog/`: raw `src`, `srcset` and `poster` attributes now go through `useBaseUrl`, matching what `docs/about/contributing.mdx` already does.
- **`src/theme/NotFound/Content/index.js`**: the chirpy image on the 404 page, the only instance outside the content directories.
- **`docusaurus.config.js`**: `baseUrl` is hoisted to a const so the footer copyright, which is a raw HTML string, can interpolate it for its Legal link. The const reads `DOCS_BASE_URL`, so a build can be served under a path without editing the config.
- **`scripts/sync-apple-docs.js`**: the generator was itself emitting raw absolute paths, so the five Apple pages are fixed at the source rather than by hand. It now wraps `src`, `srcset` and `poster` in `useBaseUrl` and adds the import, leaving Markdown syntax and fenced code samples alone. Those pages pick the change up on the next sync.

## Why did you change it

Nothing is broken on meshtastic.org, because `baseUrl` is `/` there and these paths happen to be correct. They stop being correct the moment the site is served from anywhere else: a preview under a path, a reverse proxy mounting the docs below `/docs`, a GitHub Pages project site, or an offline export. The site currently only works at a domain root, and nothing in the code says so.

The generator is the part worth a second look. Its Markdown branch was already fine, since `alt` gets `baseUrl` applied, but its `` branch wrote absolute paths, and its skip rule (`(?!\/img\/)`) meant references that were already absolute were passed over rather than corrected. Every weekly sync therefore reintroduced the pattern, and a hand fix to those pages would have lasted until Sunday.

## Anything to note

- **No rendered change at the current `baseUrl` of `/`.** `useBaseUrl("/img/x")` returns `/img/x` when `baseUrl` is `/`, so today's output is unchanged. The Vercel preview is the check for that.
- `useBaseUrl` will not double-prefix: it early-returns when the URL already starts with `baseUrl`, and leaves anchors and external URLs alone. That is what makes the sweep safe to apply mechanically.
- The `.md` files under `docs/software/apple/` take an `import` because `markdown.format` is unset, so Docusaurus 3 parses `.md` as MDX. They are treated the same as the `.mdx` files.
- Two pre-existing MDX lint complaints in `docs/community/enclosures/techo.mdx` (an HTML `<!-- -->` comment) are untouched and present on `master`. `pnpm run lint:mdx` passes.
- One raw `<a href="/documents/blog/Opposition_Letter.pdf">` in a 2024 blog post is included in the same sweep.
- Separately from this PR: the 404 image lives in the `static/design` submodule, so a clone that has not run `git submodule update --init` renders it broken no matter what `baseUrl` says.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved image, video, audio, document, and footer link resolution when the documentation site is hosted under a subpath.
  - Updated blog posts, documentation pages, and the 404 page to display assets correctly across different deployment URLs.

- **Improvements**
  - Added configurable site base URL support through the deployment environment.
  - Updated Apple documentation synchronization to automatically handle asset paths and required imports.

### DIFF
--- a/docs/hardware/devices/community-supported/lilygo/tbeam/buttons.mdx
+++ b/docs/hardware/devices/community-supported/lilygo/tbeam/buttons.mdx
@@ -5,6 +5,8 @@ sidebar_label: Buttons
 sidebar_position: 1
 ---
 
+import useBaseUrl from "@docusaurus/useBaseUrl";
+
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 
@@ -29,9 +31,9 @@ values={[
   - **Double press:** Sends an adhoc ping of the device's position to the network.
   - **Triple press:** Disables the device's GPS. Repeat to re-enable. (_This will be indicated on both information screen pages on the device's display as shown below_)
 
-<img src="/img/hardware/GPS-disabled.webp" alt="T-Beam display showing the GPS disabled indicator" width="200" align="center-left" />
+<img src={useBaseUrl("/img/hardware/GPS-disabled.webp")} alt="T-Beam display showing the GPS disabled indicator" width="200" align="center-left" />
 <img
-  src="/img/hardware/GPS-disabled-by-button.webp"
+  src={useBaseUrl("/img/hardware/GPS-disabled-by-button.webp")}
   alt="T-Beam second info screen showing the GPS disabled indicator"
   width="200"
   align="center-right"
--- a/blog/2024/August/meshtastic-opposition-to-nextnav-proposed-changes.mdx.mdx
+++ b/blog/2024/August/meshtastic-opposition-to-nextnav-proposed-changes.mdx.mdx
@@ -9,6 +9,8 @@ hide_table_of_contents: false
 image: "/design/web/social-preview-1200x630.png"
 ---
 
+import useBaseUrl from "@docusaurus/useBaseUrl";
+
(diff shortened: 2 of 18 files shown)


rcarteraz:
This is the dumbest change ever. I hate you.

rcarteraz (approved):

docs/hardware/devices/community-supported/lilygo/tbeam/buttons.mdx
```diff
@@ -29,9 +31,9 @@ values={[
   - **Double press:** Sends an adhoc ping of the device's position to the network.
   - **Triple press:** Disables the device's GPS. Repeat to re-enable. (_This will be indicated on both information screen pages on the device's display as shown below_)
 
-<img src="/img/hardware/GPS-disabled.webp" alt="T-Beam display showing the GPS disabled indicator" width="200" align="center-left" />
+<img src={useBaseUrl("/img/hardware/GPS-disabled.webp")} alt="T-Beam display showing the GPS disabled indicator" width="200" align="center-left" />
```
rcarteraz, line 34: This is going to conflict with my open PR for buttons. But please continue on your crusade Tommy boy!

NomDeTom: no pain, no gain

PR #2678, opened 2026-09-11
Add glossary tooltips to the docs

rcarteraz:
## What

Terms now link and show their definition on hover wherever they appear, so a reader no longer has to leave the page to look one up. Raised after a session where terms were missing from the glossary and the only way to read a definition was to navigate to it.

The glossary page keeps its URL at `/docs/terms/` and is driven by `glossary/glossary.json`, replacing the hand-maintained definition list. Adding a term is now a one-line edit to that file.

## Screenshots

Hovering a term in running prose, on the introduction page:

Introduction page with LoRa underlined in the opening sentence and a tooltip open beneath it giving the definition

Terms are picked up inside tables too, shown here in dark mode on the telemetry page, where ADC was the only unexplained term in the sensor list:

Sensor table in dark mode with ADC underlined in a cell and a tooltip open showing its definition

The glossary page, with live search and letter navigation, inside the normal docs layout:

Glossary of Terms page with the docs sidebar alongside it, a search box, A to W letter navigation and the first term entries

## Terms

81 in total: the 40 from the old page, plus radio and hardware jargon, Meshtastic-specific ones such as MUI, TAK and ATAK, and the IP enclosure ratings. ADC, VOC and NOx come from the review on #2659.

Definitions come from the docs and firmware where those describe a term. Standard electronics abbreviations (USB, UART, I2C, SPI, SWD, NFC, OLED, LED, MOSFET, MCU, ADC) rest on general usage, since the docs never expand them. The IP ratings come from IEC 60529 for the same reason.

Ordinary words and platform names are defined but excluded from auto-linking, so only jargon and acronyms are marked in prose.

## How the page is rendered

`docs/terms/index.mdx` is a normal doc that renders the term list from the plugin's data, rather than letting the plugin register its own route. A plugin route renders outside the docs layout, so opening the glossary dropped the sidebar and the reader's place in the tree — the problem in mcclowes/docusaurus-plugin-glossary#105. As a doc it keeps the sidebar, the version badge and the edit link.

## Trade-off worth reviewing

The eleven terms that previously linked out to their configuration or hardware pages (Channel, CLI, ESP32, Mesh, Module, nRF52, Protobuf, PSK, PKC, Repeater, Router) lose that inline link, because definitions are plain strings. Their prose definitions are unchanged. The plugin's `main` branch has since gained a per-term `documentation.path`, so a future release may allow those links back.

## Plugin patch

Three changes are carried as a pnpm patch in `patches/`:

- Tooltips opened underneath the sticky navbar. Placement measured available space from the viewport top rather than the navbar's bottom edge, and Docusaurus wraps content in ancestors using `isolation: isolate`, so the tooltip's `z-index` could never win.
- Terms matched inside words. The plural handling accepted a match that had already failed the leading word-boundary test, so "enables" rendered as "ena[bles]" linked to BLE.
- `routePath: null` suppresses the plugin's own page. The plugin stays registered so its data reaches the tooltips and the glossary page, but without this the two pages collide at the same path and Docusaurus warns about duplicate routes.

The first two are reported upstream with fixes: mcclowes/docusaurus-plugin-glossary#152 and mcclowes/docusaurus-plugin-glossary#153.

Acronym expansion is disabled. It rewrote authored prose, and its guard against expanding an already expanded acronym only matches the long form exactly, so "as a Protocol Buffer (protobuf)" became "Protocol Buffer (Protocol Buffers (protobuf))".

## Also

- `--host` and `--ble` in two Python CLI pages are now in code formatting, matching the `--noproto` in the same sentences. Bare in prose, the hyphen read as a word boundary and "ble" linked.
- The glossary entry spelled the Nordic family `NRF52`. The design standards list `nRF52840` as canonical, and the docs already favour `nRF52` 71 to 6.
- `remark-deflist` stays, since the frozen 2.7 snapshot still uses that syntax.

## Note for maintainers

Editing `glossary.json` alone does not invalidate the MDX build cache. A local rebuild needs `node_modules/.cache` and `.docusaurus` cleared before new terms appear. CI is unaffected, building from a fresh checkout.

## Testing

`pnpm run lint:mdx` and `pnpm run build` pass, with no duplicate-route warning. Verified in a browser that terms link and show tooltips in prose and inside tables, that the glossary page renders with the docs sidebar and searches, that its anchors match the links terms point at, that tooltips near the top of a page stay clear of the navbar, and that the frozen 2.7 snapshot is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added an interactive glossary with alphabetical browsing, search, term counts, definitions, and related-term links.
  * Added glossary entries for NOx and VOC.
  * Glossary terms now provide styled tooltips throughout documentation.

* **Documentation**
  * Improved formatting for connection options in Python CLI documentation.
  * Added responsive styling and dark-theme support for glossary pages.

* **Bug Fixes**
  * Improved glossary tooltip positioning when a sticky header is present.
  * Improved matching of glossary terms and plural forms.


NomDeTom (approved):
I love it

PR #2684, opened 2026-09-11
fix: honour baseUrl in raw HTML asset references

NomDeTom:
useBaseUrl() fixes for raw HTML img/source tags, plus the sync-apple-docs.js generator.


NomDeTom:
Closing — superseded by #2674/#2675, which already fix this via the base-url-assets remark plugin. Rebasing this branch would reintroduce the old manual-useBaseUrl approach on top of that.

PR #2688, opened 2026-09-12
Add raster favicon fallbacks and a web manifest

NomDeTom:
## Why

Since #2204 the site's only icon is `<link rel=icon href=img/logo.svg>`. That is fine in Chrome, Edge and Firefox, but:

- **Safari** does not render SVG favicons — the tab shows a blank icon.
- **iOS "Add to Home Screen"** has no `apple-touch-icon`, so it uses a screenshot of the page.
- **Android "Add to Home Screen"** has no manifest, so the site is not installable and the shortcut gets a low-resolution icon.

Before #2204 the favicon was `design/web/favicon.ico`, which covered the Safari case; the home-screen cases were never wired up.

## What

- `headTags` in `docusaurus.config.js` linking the existing icon set from the `static/design` submodule (`design/web/`, unchanged since 2021): 16 px and 32 px PNG icons, `apple-touch-icon` (180 px), Safari `mask-icon`, a manifest link and `theme-color`. `favicon: "img/logo.svg"` stays as the primary; browsers that support it prefer the SVG.
- `static/site.webmanifest`, the docs site's own. The design repo's `web/site.webmanifest` points at `/android-chrome-*.png`, which does not exist under this site; this one points at `/design/web/…` and names the docs site.

All hrefs use the config's `baseUrl` const, so they resolve correctly when the site is built under a prefix.

## Not done

- Not switched to `@docusaurus/plugin-pwa`. It would inject the same tags via `pwaHead` but also registers a service worker, which is a behaviour change worth its own discussion. Moving these entries into `pwaHead` later is mechanical.
- Draft because I have not run a full `docusaurus build` with this change yet.


rcarteraz:
> * **Safari** does not render SVG favicons — the tab shows a blank icon.

Are you referring to Safari on iOS? Because it definitely displays on macOS. 
Screenshot 2026-09-13 at 5 31 18 PM

Actually, no, just checked and that works on iOS too. 

image

NomDeTom:
> Are you referring to Safari on iOS? Because it definitely displays on macOS. 

You must be using the latest hotness. 

https://caniuse.com/link-icon-svg

PR #2691, opened 2026-09-13
Glossary and terms update

NomDeTom:
- Adds glossary entries for **Frequency Slot** and **Duty Cycle**, and lists Frequency Slot as a related term of Channel.
- The glossary route path now carries `baseUrl`, so term links resolve when the site is built under a path prefix. No change for the live site.

## Summary by CodeRabbit

- **New Features**
  - Added glossary entries for “Duty Cycle” and “Frequency Slot.”
  - Linked “Channel” terminology to the new “Frequency Slot” entry.

- **Bug Fixes**
  - Corrected glossary links for deployments hosted under a non-root URL.

- **Documentation**
  - Clarified “Frequency Slot” terminology and selection behavior.
  - Marked the radio-frequency meaning of “Channel” as deprecated and identified “Frequency Slot” as the preferred term.


rcarteraz:
So we now have to make sure everything has baseurl so it specifically builds on **your** machine? That's quite a maintainability ask, no?

NomDeTom:
> So we now have to make sure everything has baseurl so it specifically builds on **your** machine? That's quite a maintainability ask, no?

From the module above that:
```
// Raw HTML in the config and in MDX bypasses Docusaurus link handling, so anything
// built by hand has to prefix this itself. Overridable for builds served under a
// path rather than at a domain root. Docusaurus gives the value it is handed both a
// leading and a trailing slash, so do the same here or the two disagree.
const baseUrl = `/${process.env.DOCS_BASE_URL ?? ""}/`.replace(/\/{2,}/g, "/");
```

PR #2719, opened 2026-09-20
docs: add a bootloader-mode section for single-button nRF52 boards

### LINKED ISSUE #2718: nRF52 flashing pages have no bootloader-mode entry for single-button boards (T1000-E, X1)
The nRF52 flashing pages tell the reader to double-press the reset button to reach bootloader mode: `drag-n-drop.mdx` step 3, `nrf52-erase.mdx` step 3, and `update-booloader.mdx` step 3. The SenseCAP Card Tracker T1000-E and SenseCAP MeshTracker X1 have one button and no reset button, so that gesture cannot reach DFU on either board. OTAFIX 2.5 (2026-09-08) gave both boards a hold-to-enter path: hold the button while the device starts for 3 seconds to mount the UF2 drive, eject the drive to leave. Nothing on the site describes it, and the T1000-E and X1 device pages carry no bootloader-mode text at all.

`update-booloader.mdx` also describes Seeed's stock-bootloader gesture for the T1000-E without saying which bootloader it applies to, and `nrf52-erase.mdx` links the factory-erase file at the 2.4 release rather than the latest.

Fix: one home for entering and leaving bootloader mode on `drag-n-drop.mdx`, linked from the other two pages and from the two device pages.

jamesarich:
## Linked issue
Closes #2718

## What did you change
- New page `nrf52/entering-bootloader-mode.mdx`, the nRF52 counterpart of the STM32 one, organized by method: reset button, Meshtastic client (`meshtastic --enter-dfu`, Android's USB File Transfer), devices with a single button (T1000-E and MeshTracker X1 on OTAFIX 2.5 or later: hold 3 s through boot with USB connected, eject to leave, unplugging doesn't), and checking the bootloader version. Sidebar renumbered so it sits first.
- `drag-n-drop.mdx`, `nrf52-erase.mdx` and `update-booloader.mdx` link that page from their intro and step 3. The erase file link is `releases/latest/download/`. Seeed's own T1000-E gesture stays, scoped to the bootloader Seeed ships. Method 3 says the device comes back in Bluetooth OTA wait after a serial DFU of the bootloader package on OTAFIX 2.0 or later. `caution` becomes `warning` on both pages.
- `card-tracker.mdx` and `meshtracker-x1.mdx` gain a Flashing section with the numbered steps; the T1000-E button list points at it and its `Functionality` heading level is fixed.

## Why did you change it
OTAFIX 2.5 (2026-09-08) gave both single-button boards a hold-to-enter path (meshtastic/Adafruit_nRF52_Bootloader_OTAFIX#45, #47, #48). Every nRF52 flashing page still said double-press reset, which cannot reach DFU on either board, and the device pages said nothing.

Measured on hardware in those PRs: T1000-E hold through a soft reboot mounts the drive about 4 s after the reset, and eject exits (#48, #47); X1 hold and eject (#45). From source only: powering the device on with the button held (step 2 on both device pages), which relies on the `DFU_MAGIC_SKIP` fall-through in `main.c`. Needs a T1000-E and an X1 before merge.

`pnpm lint:mdx` passes on the changed directories. `pnpm build` passes with `onBrokenLinks: throw`; its SSG warnings are all under `versioned_docs/2.7`, untouched here.

## Screenshots
None, text only.

## Summary by CodeRabbit

- **Documentation**
  - Added a dedicated nRF52 bootloader-mode guide covering reset-button entry, client-command entry, USB transfers, exit behavior, bootloader versions, and factory-erase support.
  - Updated firmware flashing, factory-erase, bootloader-update, and OTA instructions with clearer navigation, current download links, and device-specific procedures.
  - Added SenseCAP Card Tracker and MeshTracker X1 flashing guidance, including OTAFIX requirements and USB firmware installation.
  - Reorganized the nRF52 flashing documentation sidebar for improved navigation.


docs/getting-started/flashing-firmware/nrf52/update-booloader.mdx
```diff
@@ -85,6 +85,8 @@ adafruit-nrfutil --singlebank --touch 1200 --verbose dfu serial --package <downl
 
 5. Once the process finishes, the update is complete. Your device is now running the latest bootloader version and you can proceed with [flashing the firmware](/docs/getting-started/flashing-firmware/nrf52/drag-n-drop/).
 
+On OTAFIX 2.0 or later, flashing the bootloader package this way clears the bootloader's record of the installed firmware, so the device comes back waiting for a Bluetooth OTA update rather than starting the firmware: no USB drive and no serial port appear. That is expected, not a failure. Double-press the reset button for the UF2 drive and copy the firmware across, or complete a [Bluetooth OTA update](/docs/getting-started/flashing-firmware/nrf52/ota/).
```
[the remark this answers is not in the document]
jamesarich: Reworded to link the bootloader-mode section instead of naming double-press. Not claiming a single-button route out of OTA wait: `--enter-dfu` needs a running firmware, and nobody has measured what a T1000-E does there.

docs/getting-started/flashing-firmware/nrf52/drag-n-drop.mdx
```diff
@@ -10,6 +10,25 @@ import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 import Link from "@docusaurus/Link";
 
+## Bootloader mode
```
rcarteraz: This page covers nRF52, RP2040 and RP2350, and there's an `### RP2040` section further down that uses BOOTSEL. A section opening "Every nRF52 device runs a small program..." above `## Flash Firmware` therefore puts chip-specific background ahead of the task for everyone, including readers it doesn't apply to.

The site already has a home for exactly this content. `stm32/entering-bootloader-mode.mdx` is `sidebar_position: 1`, opens with a short "most convenient first" list, then carries one H2 per method (Meshtastic Client, BOOT0, SWD) and no table. A matching `nrf52/entering-bootloader-mode.mdx` would fix three things at once: the placement, the table below, and the deep anchor links, since `nrf52-erase.mdx`, `update-booloader.mdx`, `card-tracker.mdx` and `meshtracker-x1.mdx` all now point at `drag-n-drop/#bootloader-mode`, an anchor inside a task page.

Organized by method rather than by device, which is also what the STM32 page does:

- **Reset button**: double-press, press once to leave
- **Meshtastic client**: `meshtastic --enter-dfu`, plus Android's **USB File Transfer**
- **Devices with a single button**: T1000-E and X1, OTAFIX 2.5 or later, eject to leave
- **Checking your bootloader version**: `INFO_UF2.TXT`, linking the update page

Step 3 here then stays one line and links a page instead of an anchor. The sidebar renumber would be the new page 1, `drag-n-drop` 2, `ota` 3, `nrf52-erase` 4, `convert-rak4631r` 5, `update-booloader` 6, `swdio` 7. (`index.mdx` and `ota.mdx` both sit at 2 today, which is pre-existing and not yours to fix here.)

docs/getting-started/flashing-firmware/nrf52/drag-n-drop.mdx
```diff
@@ -10,6 +10,25 @@ import Tabs from "@theme/Tabs";
...
+| SenseCAP Card Tracker T1000-E and SenseCAP MeshTracker X1 on the [Meshtastic OTAFIX bootloader](https://github.com/meshtastic/Adafruit_nRF52_Bootloader_OTAFIX) 2.5 or later | With the USB cable connected, hold the button while the device starts and keep holding it for 3 seconds. The drive appears a few seconds later. Power the device on with the button held, or reboot it from a client while holding the button. | Eject the drive from your computer. Unplugging doesn't leave bootloader mode: the device stays in it on battery, and plugging it back in only mounts the drive again. |
+
+The T1000-E and X1 have one button and no reset button. A short press while the device starts, or a hold while the firmware is running, doesn't enter bootloader mode. On the T1000-E a long press while the firmware runs powers the device off; press and hold the button again to power it on and reach the bootloader.
```
rcarteraz: The table is doing prose's job. The second row carries three sentences in one cell and two in the next, which 11.10 addresses directly: "Keep structure simple: deeply nested lists and multi-paragraph table cells are hard to navigate without sight."

11.5 says prefer a table "wherever content has a repeating shape". Two rows, one of which means "everything else", is a fork rather than a shape, and the same section asks for front-loading, the rule before its exceptions. As it stands a reader on a RAK board parses both rows to learn they just need a double-press. The first column doesn't help either, since it's a 20-word device name carrying a link and a version qualifier, so the table can't be scanned by its key column.

It's also unlike the rest of the site: nothing under `docs/hardware/` tables button gestures or bootloader entry, and the only table with button rows anywhere is a pin map in `nucleo-wl55jc.mdx`. A two-row table invites a third row per device; prose stating one rule and one exception doesn't.

Keeping the USB-cable requirement as prominent as you have it is right, for the `uf2_dfu` timeout reason in the summary. This prose works unchanged on a separate page, so it isn't wasted if you take the split.

I've left the "power the device on with the button held" sentence out pending the hardware check, since on its own the paragraph still reads correctly without it.

```suggestion
Most nRF52 devices enter bootloader mode when you double-press the reset button within about half a second. Press the reset button once to leave it.

The SenseCAP Card Tracker T1000-E and SenseCAP MeshTracker X1 have one button and no reset button, so the double-press doesn't apply. On the [Meshtastic OTAFIX bootloader](https://github.com/meshtastic/Adafruit_nRF52_Bootloader_OTAFIX) 2.5 or later, connect the USB cable, then hold the button while the device starts and keep holding it for 3 seconds. The drive appears a few seconds later. A short press while the device starts, or a hold while the firmware is running, doesn't reach the bootloader.

To leave bootloader mode without flashing, eject the drive. Unplugging doesn't leave bootloader mode: the device stays in it on battery, and plugging it back in only mounts the drive again.
```

docs/hardware/devices/seeed-studio/sensecap/card-tracker.mdx
```diff
@@ -43,6 +43,7 @@ Currently, LR1110 radios are unable to receive Meshtastic packets from the older
   - **Double press:** Sends an adhoc ping of the device's position to the network.
   - **Triple press:** Enables/Disables the GPS Module on demand.
   - **Quadruple press:** Temporarily mute or unmute device.
+  - **Hold for 3 seconds while the device starts:** Enters [bootloader mode](/docs/getting-started/flashing-firmware/nrf52/drag-n-drop/#bootloader-mode) on the Meshtastic OTAFIX bootloader 2.5 or later. A hold while the firmware is running doesn't.
```
rcarteraz: Entering DFU is a maintenance procedure sitting in a list whose other four entries are runtime actions: ping, GPS toggle, mute. Readers looking for it search "T1000-E DFU", not "buttons".

It also drops the USB-cable requirement the table states, and that one matters: `main.c` restarts into the app if USB isn't enumerated within 3 seconds, so without the cable the hold appears to do nothing at all. "While the device starts" is circular here too, because on a T1000-E the only way to start the device is that same button, so the reader needs "press and hold to power on, and keep holding".

`Flashing` is what 21 other device pages call this section, for instance `heltec-automation/mesh-node/index.mdx`, and it already sits between Features or Functionality and Resources. `###` here to match this page's `### Features` and `### Resources`.

Step 2 below is the path your PR body flags as source-only. Worth confirming on hardware before merge, as noted in the summary.

```suggestion
  - **Hold while the device starts:** Enters bootloader mode for a firmware update. See [Flashing](#flashing).

### Flashing

The T1000-E has one button and no reset button, so the double-press that reaches bootloader mode on most nRF52 devices doesn't apply. This needs the [Meshtastic OTAFIX bootloader](https://github.com/meshtastic/Adafruit_nRF52_Bootloader_OTAFIX) 2.5 or later.

1. Connect the device to your computer with a USB data cable.
2. Press and hold the button to power the device on, and keep holding it for 3 seconds.
3. The device mounts as a USB drive a few seconds later.
4. Copy the firmware `.uf2` file onto the drive, as described in [Drag & Drop](/docs/getting-started/flashing-firmware/nrf52/drag-n-drop/).

To leave bootloader mode without flashing, eject the drive.

On an earlier bootloader, or to skip the button entirely, connect the device over USB and run `meshtastic --enter-dfu`.

```

docs/hardware/devices/seeed-studio/sensecap/meshtracker-x1.mdx
```diff
@@ -32,7 +32,7 @@ The SenseCAP MeshTracker X1 is a card-sized IP66-rated tracker for Meshtastic, b
 - IP66-rated enclosure.
 - Vibration motor for alerts.
 - Barometric pressure sensor.
-- User button and RGB status LED.
+- User button and RGB status LED. Holding the button for 3 seconds while the device starts enters [bootloader mode](/docs/getting-started/flashing-firmware/nrf52/drag-n-drop/#bootloader-mode) on the Meshtastic OTAFIX bootloader 2.5 or later; there is no reset button.
```
rcarteraz: This bullet now carries three things: a hardware feature, a procedure, and the no-reset-button fact. It's also invisible to anyone searching "X1 DFU", because it isn't a heading.

Moving the procedure into its own section (next comment) leaves this as a spec bullet again. The no-reset-button fact is worth keeping here, since that genuinely is a hardware property.

```suggestion
- User button and RGB status LED. There's no reset button.
```

docs/hardware/devices/seeed-studio/sensecap/meshtracker-x1.mdx
```diff
@@ -32,7 +32,7 @@ The SenseCAP MeshTracker X1 is a card-sized IP66-rated tracker for Meshtastic, b
...
 - No screen, so the device is configured from a client.
```
rcarteraz, line 37: The X1 has no Flashing or Functionality section at all today, which is why the gesture ended up in a Features bullet; there's nowhere else for it to go. `##` here to match this page's other headings, and placed before Resources like the other device pages.

The content is the same as the T1000-E's, which is fine, since it's the same gesture on the same bootloader scheme.

Separately, and out of scope for this PR: the X1's button gestures aren't documented anywhere, so the page could use a `## Functionality` block matching its sibling `card-tracker.mdx`.

```suggestion
- No screen, so the device is configured from a client.

## Flashing

The MeshTracker X1 has one button and no reset button, so the double-press that reaches bootloader mode on most nRF52 devices doesn't apply. This needs the [Meshtastic OTAFIX bootloader](https://github.com/meshtastic/Adafruit_nRF52_Bootloader_OTAFIX) 2.5 or later.

1. Connect the device to your computer with a USB data cable.
2. Press and hold the button to power the device on, and keep holding it for 3 seconds.
3. The device mounts as a USB drive a few seconds later.
4. Copy the firmware `.uf2` file onto the drive, as described in [Drag & Drop](/docs/getting-started/flashing-firmware/nrf52/drag-n-drop/).

To leave bootloader mode without flashing, eject the drive.

On an earlier bootloader, or to skip the button entirely, connect the device over USB and run `meshtastic --enter-dfu`.

```

docs/getting-started/flashing-firmware/nrf52/nrf52-erase.mdx
```diff
@@ -47,7 +47,7 @@ Bootloaders without a `Factory-Erase:` line in `INFO_UF2.TXT`, which includes ev
 3. With the [Meshtastic CLI](/docs/software/python/cli/) installed, open your Terminal/Console and enter the command: `meshtastic --noproto`.
 4. You should see the message: `Formatting... Done`.
 
-:::caution
+:::warning
 The erase file must match your SoftDevice version: the wrong one can overwrite part of the SoftDevice and leave the device needing a hardware programmer to recover. If you copied `meshtastic_factory_erase.uf2` and nothing happened, your bootloader does not support it yet: use the steps above, or [update the bootloader](/docs/getting-started/flashing-firmware/nrf52/update-nrf52-bootloader) first and check `INFO_UF2.TXT` again.
```
rcarteraz: Minor, non-blocking, while you're in this hunk. Two things on this line: "the steps above" points at another part of the page, which 11.5 asks you not to do, and the bootloader link is missing the trailing slash that `drag-n-drop.mdx` uses for the same target.

```suggestion
The erase file must match your SoftDevice version: the wrong one can overwrite part of the SoftDevice and leave the device needing a hardware programmer to recover. If you copied `meshtastic_factory_erase.uf2` and nothing happened, your bootloader does not support it yet: use the preceding steps, or [update the bootloader](/docs/getting-started/flashing-firmware/nrf52/update-nrf52-bootloader/) first and check `INFO_UF2.TXT` again.
```

rcarteraz (changes requested):
The fix here is real and the source backs it. I checked OTAFIX `master`: `BUTTON_DFU_HOLD` in `src/main.c` sets `uf2_dfu = true` explicitly rather than falling through to the OTA default, so the hold really does give the UF2 drive rather than BLE OTA; `BUTTON_DFU_HOLD_MS` is 3000 on both `t1000_e` and `mesh_tracker_x1`; and both boards define `UF2_EXIT_ON_EJECT`, so "eject to leave" is right for both. `meshtastic --enter-dfu` and the Android **USB File Transfer** label both check out too, the latter at `docs/software/android/user/firmware.md:39`, which may be worth linking rather than restating. The `caution` to `warning` fixes, the `releases/latest/download/` link, and scoping Seeed's own gesture to the bootloader Seeed ships are all good changes.

So my objection is structural rather than factual, and it's really one ask: this material wants its own page. `drag-n-drop.mdx` covers nRF52, RP2040 and RP2350, so a section opening "Every nRF52 device..." placed above `## Flash Firmware` puts chip-specific background ahead of the task for every reader of the page, including the ones it doesn't apply to. The site already has the pattern in `stm32/entering-bootloader-mode.mdx`, and a matching `nrf52/entering-bootloader-mode.mdx` would fix the placement, let the table become prose, and turn four deep links into `drag-n-drop/#bootloader-mode` into links to a page. The inline comments carry the specifics, with suggestions where they're applicable.

Two things that don't sit on changed lines. First, the "power the device on with the button held" clause: your PR body flags it as source-only, and I agree it should be confirmed on a T1000-E and an X1 before this merges. It isn't a doubt about the mechanism, since `main.c:249` handles that boot deliberately and the comment there names the System OFF wake case, but it's the only entry path for a device sitting powered off, so it's the sentence most worth being certain about. Related, and new: `main.c` gives the `uf2_dfu` path a 3-second USB enumeration timeout, "If USB is not enumerated in 3s (eg. because we're running on battery), we restart into app". That makes "with the USB cable connected" load-bearing rather than incidental. The table states it; both device-page bullets drop it, which is the one place this PR is actually wrong rather than just cramped.

Second, `card-tracker.mdx:39` has `## Functionality` sitting between `### Features` and `### Resources`, an out-of-order heading level against 11.5 and 11.10. It's outside the diff, but you're editing inside that section, so it's cheap to carry along.

docs/getting-started/flashing-firmware/nrf52/drag-n-drop.mdx
jamesarich: Split out as `nrf52/entering-bootloader-mode.mdx`, organized by method as on the STM32 page; sidebar renumbered as you listed, the anchor links are page links now.

docs/getting-started/flashing-firmware/nrf52/drag-n-drop.mdx
jamesarich: Table gone, your prose used on the new page; the power-on sentence stays out of it pending the hardware check.

docs/hardware/devices/seeed-studio/sensecap/card-tracker.mdx
jamesarich: Taken as suggested: bullet points at a new `### Flashing` with the USB cable as step 1, plus a link to the methods page.

docs/hardware/devices/seeed-studio/sensecap/meshtracker-x1.mdx
jamesarich: Bullet is back to the spec line with the no-reset-button fact.

docs/hardware/devices/seeed-studio/sensecap/meshtracker-x1.mdx
jamesarich: Added `## Flashing` before Resources as suggested; the X1 Functionality block is a separate change.

docs/getting-started/flashing-firmware/nrf52/nrf52-erase.mdx
jamesarich: Both taken.

docs/hardware/devices/seeed-studio/sensecap/card-tracker.mdx
```diff
@@ -36,13 +36,27 @@ Currently, LR1110 radios are unable to receive Meshtastic packets from the older
...
 - **User/Program Button (face):**
   - **Long press:** Will signal the device to shutdown after 5 seconds.
   - **Double press:** Sends an adhoc ping of the device's position to the network.
   - **Triple press:** Enables/Disables the GPS Module on demand.
   - **Quadruple press:** Temporarily mute or unmute device.
+  - **Hold while the device starts:** Enters bootloader mode for a firmware update. See [Flashing](#flashing).
+
+### Flashing
+
+The T1000-E has one button and no reset button, so the double-press that reaches bootloader mode on most nRF52 devices doesn't apply. This needs the [Meshtastic OTAFIX bootloader](https://github.com/meshtastic/Adafruit_nRF52_Bootloader_OTAFIX) 2.5 or later.
```
rcarteraz, line 50: For the T1000 I would also link to Seeed's wiki on entering boot loader making a note that if they haven't flashed Meshtastic's bootloader then they will need to follow that process or update the boot loader.

rcarteraz (approved):
Left a comment on a single change I think should be added otherwise this looks good.

docs/hardware/devices/seeed-studio/sensecap/card-tracker.mdx
jamesarich: Added: Seeed guide linked for the stock bootloader, with the bootloader update as the alternative.

